PR #22, opened 2020-08-30
Cleanup and bump appVersions

onedr0p:

Bumping versions to their latest releases and updating the chart.yaml's to include the actual app versions (not docker image version)

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/zwave2mqtt/values.yaml
+++ b/charts/zwave2mqtt/values.yaml
@@ -7,7 +7,7 @@ strategyType: Recreate
 
 image:
   repository: robertslando/zwave2mqtt
-  tag: 3.0.1
+  tag: 4.0.3
   pullPolicy: IfNotPresent
 
 # timezone to run the service as
--- a/charts/bazarr/Chart.yaml
+++ b/charts/bazarr/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v2
-appVersion: v0.8.4.2-ls72
+appVersion: v0.9.0.2
 description: Bazarr is a companion application to Sonarr and Radarr. It manages and downloads subtitles based on your requirements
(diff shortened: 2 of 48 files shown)


charts/zwave2mqtt/values.yaml
```diff
@@ -7,7 +7,7 @@ strategyType: Recreate
 
 image:
   repository: robertslando/zwave2mqtt
-  tag: 3.0.1
+  tag: 4.0.3
```
billimek, line 10: This version caused some issues with excessive logging. I suggest not updating this chart just yet.

onedr0p: This seems reasonable to me... 
image

onedr0p: Any chance your log file wasn't pruned after upgrading?

billimek: It does get pruned.  This is what I see after upgrading to 4.0.3.  It seems to get into an endless logging loop after changing the power state of a zwave switch.  I'll  log an issue over there.  Maybe it's a problem that only affects certain setups:

image

That's approx 30KB/sec log growth rate.

onedr0p: Gotcha, I'll bump this back down to 3.4.1

billimek: With that said, I'm good either way with the version you have set - it's easy to override.  I did raise https://github.com/OpenZWave/Zwave2Mqtt/issues/697 for the issue.

onedr0p: Not sure about the lint issue, maybe because this is so large, but we can get it merged as is.

PR #33, opened 2020-09-05
[media-common] New Chart

DirtyCajunRice:
Initially has radarr/sonarr/lidarr/ombi/organizr/tautulli. Can add bazarr, nzbget, and sabnzbd in the same fashion.
original PR: https://github.com/billimek/billimek-charts/pull/329


.idea/k8s@home charts.iml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
bjw-s, line 1: Might need to `.gitignore` this, as not everyone will be using a Jetbrains IDE.

bjw-s:
Looks great! Is there a reason why you put the actual services under the `media-common` folder instead of the top level `charts` folder? I think that could increase findability (is that a word? 🤔)

.idea/k8s@home charts.iml
DirtyCajunRice: Good catch! Didnt even notice.

DirtyCajunRice:
> Looks great! Is there a reason why you put the actual services under the `media-common` folder instead of the top level `charts` folder? I think that could increase findability (is that a word? )

So i was talking with @onedr0p  and because they are literally just shells with almost no information i had a hard time swallowing how much bulk that puts in the top level for what it actually was. If yall think it should be top level i can move them.

DirtyCajunRice:
After checking, we can move them to the top as replacements for their former versions, but they reference a chart that doesnt exist yet (media-common) and will break the releaser if they are released at the same time as it. 

In addition, the releaser's logic does not allow nesting like that so they have to go to the base level anyway :(.

DirtyCajunRice:
Ok. MVP is ready. Steps would be:

- [ ] merge this to allow the common chart to exist (requirement for all other charts that depend on this)
- [ ] create new MR that replaces each top level chart with the one in this folder, incrementing them to the next major version as to signify a breaking change
- [ ] have someone who uses them add bazarr, nzbget, and sabnzbd

carpenike (approved):
LGTM

carpenike (approved):

PR #44, opened 2020-09-11
[autoscan] Inital Commit

dcplaya:
Autoscan replaces the default Plex and Emby behaviour for picking up 
file changes on the file system. Autoscan integrates with Sonarr, Radarr 
and Lidarr (with Google Drive coming soon!) to fetch changes in near 
real-time without relying on the file system.

Utilizes media-common since it shares the same mounts as the *arr services and just needs a config volume. 

#### Checklist
- [x] Chart Version bumped


dcplaya:
Is this saying I need to bump the `media-common` chart version? All I added was an update to the `README`, didn't think that warranted a version bump

```
------------------------------------------------------------------------------------------------------------------------
 ✔︎ autoscan => (version: "1.0.0", path: "charts/autoscan")
 ✖︎ media-common => (version: "1.0.1", path: "charts/media-common") > Chart version not ok. Needs a version bump!
------------------------------------------------------------------------------------------------------------------------
Error: Error linting charts: Error processing charts
Error linting charts: Error processing charts
```

carpenike:
yeah that likely means you need to bump both regardless of the change. @DirtyCajunRice

dcplaya:
Do we want to keep bumping the `media-common` version each time? Chart definitely works, I have it running via my forked repo. I'll bump it if that's the course we want to take.

carpenike:
@DirtyCajunRice -- any thoughts on an approach for these types of updates?

> Do we want to keep bumping the `media-common` version each time? Chart definitely works, I have it running via my forked repo. I'll bump it if that's the course we want to take.

DirtyCajunRice:
> @DirtyCajunRice -- any thoughts on an approach for these types of updates?
> 
> 
> 
> > Do we want to keep bumping the `media-common` version each time? Chart definitely works, I have it running via my forked repo. I'll bump it if that's the course we want to take.
> 
> 
> 
> 

So if we are going by helm community standards - yes. A readme update is "technically" a change to the chart. The readme would not be reflected in installs or in repo aggregators like artifacthub without a bump.

Now the more human answer is probably "make a judgement call based on the value of the readme update... if it is enough to warrant the change it is probably worth bumping so people can see it"

finally: i think we should remove references to parent charts in the readme. you can see linked charts in artifacthub so doing this just causes cyclical updates.

onedr0p:
@dcplaya any update here, should be close this?

dcplaya:
I say close it. Devs are still working on the actual code and I have had a few times where the charts broke because something was modified on the backend (file locations, commands, ect). Just not ready for primetime.

onedr0p:
Ok sounds good :)

Issue #37, opened 2020-09-06
Chart migration to use media-common


DirtyCajunRice:
Candidate charts that would work as an umbrella templated media-common chart
- [ ] bazarr
- [ ] nzbget
- [ ] sabnzbd
- [ ] dashmachine

(are there more? will add to the list as needed)

carpenike:
Could add dashmachine if organizrr is apart of it.

bjw-s:
I'll take a crack at migrating nzbget

bjw-s:
It doesn't like we'll be able to do a 1:1 migration of nzbget because of the integrated OpenVPN container :( That may be be resolved by some sort of `additionalContainers` directive in media-common, but is that a direction we're willing to go? It will make media-common a super flexible chart template, but maybe that defeats its' purpose.

DirtyCajunRice:
@bjw-s  so check this out. The point was to be able to swiss-army-knife other stuff into it. My proposal:
Make a openvpn library chart that is added as a dependency when requested and we can add an `additionalContainers` section like you said where this will utilize that and inject it in.

bjw-s:
Now that #45 is in, the only thing I'm still wondering about is the `downloads` volume that is currently in place for nzbget (and probably the other downloader-type charts). Directions we could take this:

- direct users to use the `media` volume from _media-common_ for their downloads location
- direct users to provide their own PVC and use the additionalVolumes and additionalVolumeMounts option
- add a `downloads` volume to _media-common_ similar to `media`

I have a slight preference for option 3 in this case, but I'm curious about your take

CuBiC3D:
nzbhydra2 also fits the pattern

PR #45, opened 2020-09-11
[media-common] Add openvpn add-on

bjw-s:
Adds `media-common-openvpn` add-on to `media-common`. Lays the foundation for migrating helm charts (#37) that use a VPN sidecar to the new media-common chart.

**Note** this PR depends on #46 being merged

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/media-common/Chart.yaml
+++ b/charts/media-common/Chart.yaml
@@ -2,10 +2,16 @@ apiVersion: v2
 name: media-common
 description: Common dependancy chart for media ecosystem containers
 type: application
-version: 1.0.1
+version: 1.1.0
 keywords:
   - media-common
 home: https://github.com/k8s-at-home/charts/tree/master/charts/media-common
 maintainers:
   - name: DirtyCajunRice
     email: nick@cajun.pro
+dependencies:
+  - name: media-common-openvpn
+    repository: https://k8s-at-home.com/charts/
+    version: 1.0.0
+    condition: openvpn.enabled
+    alias: openvpn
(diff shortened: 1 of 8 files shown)


charts/media-common/Chart.yaml
```diff
@@ -2,10 +2,16 @@ apiVersion: v2
 name: media-common
 description: Common dependancy chart for media ecosystem containers
 type: application
-version: 1.0.1
+version: 1.1.0
```
DirtyCajunRice, line 5: Technically this is correct in semver bumping. Doing this _will_ cause the already created containers to desync. 
I think we need to bump the semver regex to be a minor match instead of a micro match.
E.g. `~1.0.0` to `^1.0.0`

charts/media-common/Chart.yaml
```diff
@@ -2,10 +2,16 @@ apiVersion: v2
...
 keywords:
   - media-common
 home: https://github.com/k8s-at-home/charts/tree/master/charts/media-common
 maintainers:
   - name: DirtyCajunRice
     email: nick@cajun.pro
+dependencies:
+  - name: media-common-openvpn
+    repository: https://k8s-at-home.com/charts/
+    version: 1.0.0
```
DirtyCajunRice, line 15: Lets preemptively set this to  `^1.0.0` so that we can accomodate for minor updates and micro updates of the library chart

DirtyCajunRice (changes requested):
Super awesome stuff. The 2 comments below with semver changes, and adding values to ct-values.yaml to enabled this by default (so that it can be tested) would make this perfect

charts/media-common/Chart.yaml
bjw-s: Will update the dependent charts in another PR 👍

DirtyCajunRice (approved):
LTGM

PR #53, opened 2020-09-17
[home-assistant] Added the ability to specify subPaths

gandazgul:
I mount all my configs into a volume and keep each app under subPaths. By adding the ability of specifying a subpath here I can set hass to store its settings on its own folder.

thanks

#### Checklist
- [x] Chart Version bumped


charts/home-assistant/README.md
```diff
@@ -85,7 +85,8 @@ The following tables lists the configurable parameters of the Home Assistant cha
...
 | `git.user.name`                                 | Human-readable name in the “committer” and “author” fields                                                                                                                                                                                | ``                                  |
-| `git.user.email`                                | Email address for the “committer” and “author” fields                                                                                                                                                                                     | ``                                  |
+| `git.user.email`                                | Email address for the “committer” and “author” fields                    | ``                                 |                                                      
+| `git.configSubPath`                             | An optional subPath for the git configuration volumeMount                | ``                                |
```
billimek, line 89: Should we also add this to the `values.yaml` file so folks know that it's an option that can be set?  Generally if something is a configurable value, it will have a stub (or commented stub) in `values.yaml`

charts/home-assistant/README.md
```diff
@@ -94,7 +95,8 @@ The following tables lists the configurable parameters of the Home Assistant cha
...
 | `extraEnv`                                      | Extra ENV vars to pass to the home-assistant container                                                                                                                                                                                    | `{}`                                |
-| `extraEnvSecrets`                               | Extra env vars to pass to the home-assistant container from k8s secrets - see `values.yaml` for an example                                                                                                                                | `{}`                                |
+| `extraEnvSecrets`                               | Extra env vars to pass to the home-assistant container from k8s secrets - see `values.yaml` for an example                                                                                                                                | `{}`                               |
+| `configSubPath`                                 | An optional subPath for the config volumeMount                           | ``                                |
```
billimek, line 99: Should we also add this to the `values.yaml` file so folks know that it's an option that can be set?  Generally if something is a configurable value, it will have a stub (or commented stub) in `values.yaml`

billimek (commented):
We may want to add the two new fields to `values.yaml`, @gandazgul

charts/home-assistant/README.md
gandazgul: ok I can do that.

charts/home-assistant/README.md
bjw-s: I'm not completely sure if `git` is the best key to place this under, as this setting would also be applicable when not using the `git` settings at all. We (probably?) also don't set the config `mountPath` to `git.syncPath` for similar reasons.

I understand that it needs to go somewhere, but maybe it is better suited under the `persistence` key or an entirely new key?

gandazgul: this one is specifically for the git-sync config volume mount.

gandazgul:
I moved the configSubPath setting under persistence because I think it fits better there and removed the one under git because I realized is the same volume, git gets the same hass config folder in order to sync it. 

I also added configSubPath to the values.yaml commented with an explanation of its use above. 

@billimek @bjw-s please let me know what you think now and thank you for your time reviewing.

charts/home-assistant/README.md
```diff
@@ -155,6 +156,7 @@ The following tables lists the configurable parameters of the Home Assistant cha
 | `appdaemon.image.pullPolicy`                    | Image pull policy                                                                                                                                                                                                                         | `IfNotPresent`                      |
 | `appdaemon.haToken`                             | Home Assistant API token - you need to generate it in your Home Assistant profile and then copy here                                                                                                                                      | ``                                  |
 | `appdaemon.extraEnv`                            | Extra ENV vars to pass to the AppDaemon container                                                                                                                                                                                         | `{}`                                |
+| `appdaemon.configSubPath`                       | An optional subPath for the AppDaemon container's config volume mount   | `appdaemon`                          |
```
billimek, line 159: This should probably also have an entry in the `values.yaml`.  Basically, if we make reference to a helm configuration setting in the readme or actual templates, we need to have a corresponding reference in the `values.yaml` (even if it's commented out)

billimek (commented):
`appdaemon.configSubPath` should have a reference in the `values.yaml`.  Can you take a quick look?

charts/home-assistant/README.md
gandazgul: Done

billimek (dismissed):
lgtm

gandazgul:
I tested it on my k8s using my branch as the repo source and its working great. Thank you for maintaining this chart and repo.

billimek (approved):

PR #62, opened 2020-09-24
[media-common] Support additionalPorts

bjw-s:
When migrating the qBittorrent chart to media-common I ran into the issue that I need to specify 2 ports (1 for UI, 1 for bt protocol). 
I don't think I can use `service.additionalSpec` for this because the named ports would not be exposed from the deployment/sts  

#### Checklist
- [x] Chart Version bumped


bjw-s:
No longer needed, works without named ports

PR #66, opened 2020-09-26
[media-common] Add initContainers

dcplaya:
I think this is all that is needed to get this working. I don't think I need to wrap the `initContainers:` key in an if statement. 

This is needed so I can finish getting #44 properly committed. Will probably be useful for others too

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/media-common/templates/deployment.yaml
+++ b/charts/media-common/templates/deployment.yaml
@@ -27,6 +27,8 @@ spec:
       securityContext:
       {{- toYaml . | nindent 8 }}
       {{- end }}
+      initContainers:
+      {{- include "media-common.initContainers" . | nindent 8 }}
       containers:
         - name: {{ template "media-common.fullname" . }}
           {{- with .Values.securityContext }}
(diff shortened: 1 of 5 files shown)


charts/media-common/templates/deployment.yaml
```diff
@@ -27,6 +27,8 @@ spec:
       securityContext:
       {{- toYaml . | nindent 8 }}
       {{- end }}
+      initContainers:
```
DirtyCajunRice, line 30: this will create an empty initContainers key even if not used. Preferred format is

```yaml
{{- with .Values.initContainers }}
initContainers:
  {{- include "media-common.initContainers" . | nindent 8 }}
{{- end}}

File: charts/media-common/templates/statefulset.yaml [hunk 2]
```diff
@@ -28,6 +28,8 @@ spec:
       securityContext:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      initContainers:
```
DirtyCajunRice, line 31: same as the other comment

DirtyCajunRice (changes requested):
Nice. Close!

charts/media-common/templates/deployment.yaml
dcplaya: Better than an if statement too!

DirtyCajunRice (approved):
LGTM

PR #87, opened 2020-10-07
[Adguard-home] Allow to mount secret with certs so we can set the tls …

fonsecas72:
#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/adguard-home/Chart.yaml
+++ b/charts/adguard-home/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: v0.102.0
 description: DNS proxy as ad-blocker for local network
 name: adguard-home
-version: 2.1.1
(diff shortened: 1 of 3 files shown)


charts/adguard-home/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: v0.102.0
 description: DNS proxy as ad-blocker for local network
 name: adguard-home
-version: 2.1.1
+version: 2.1.2
```
billimek, line 5: As there is a new feature being added, can you please bump the chart version to 2.2.0?

fonsecas72: yeah, makes sense. Thanks :)

billimek (approved):
lgtm

PR #95, opened 2020-10-18
media-common retro fix (1/3)

### LINKED ISSUE #73: [media-common] Breaking ingress change
Several of my deploys are failing recently and I just dug into what's going on. It appears media-common had some updates to the ingress to meet new 1.19 requirements. That change is [here](https://github.com/k8s-at-home/charts/commit/b8ae05556163dc7756e6eee2b7310d7656f427c5).

The change bumped the Minor version from 1.1.1 to 1.2.0. This being non-backwards compatible I believe it should have bumped the major version to 2.0.0 instead.

The charts repository then had [this change](https://github.com/k8s-at-home/charts/commit/12752f4018851f99694a30f041fb389e0be5f1ca) landed which switches the media-common from "~" to "^" versioning which allows Minor and Patch instead of just Patch upgrades.

The result is that all media-common charts with the 2nd change (sonarr, radarr, organizr) have a backwards incompatible change to their values.yaml. The ingress used in those charts require that the 'paths' key have a map rather than list of paths as was previously in the chart.

It's a good change, I believe the error in this is actually the first commit which should have bumped the Major version, so that anyone using the supporting charts with ~ or ^ matching wouldn't receive a backwards incompatible change. For now I've pinned my chart directly to the release from a few days ago.

I think rolling back the 1.2.0 media-common, bumping the change to 2.0 and then bumping the umbrella charts major versions would be the correct way to version this change such that existing charts are aware this is a non-backwards compatible change and won't pull it in automatically.

DirtyCajunRice:
Fixes #73 
Objective:
1. Create a final version of the charts under this major version that is pinned to the last non-breaking change in media-common
2.  Create another PR that does a major version bump of all charts + media-common using the new features, with a standard notice of the breaking change

Charts affected:
- jackett
- lidarr
- nzbget
- ombi
- organizr
- qbittorrent
- radarr
- sonarr
- tautulli

### DIFF
--- a/charts/jackett/Chart.yaml
+++ b/charts/jackett/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: v0.16.1045
 description: API Support for your favorite torrent trackers
 name: jackett
-version: 4.0.1
+version: 4.0.2
 keywords:
   - jackett
   - torrent
@@ -17,5 +17,5 @@ maintainers:
 dependencies:
   - name: media-common
     repository: https://k8s-at-home.com/charts/
-    version: ^1.0.0
+    version: 1.1.1
     alias: jackett
--- a/charts/lidarr/Chart.yaml
+++ b/charts/lidarr/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: lidarr
 description: Looks and smells like Sonarr but made for music
 type: application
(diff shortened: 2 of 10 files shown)


onedr0p:
Closed in favor of https://github.com/k8s-at-home/charts/pull/109

PR #104, opened 2020-10-27
fix stable repo

michaelkoetter:
Fixes the "Release Charts" Github workflow.

https://helm.sh/blog/new-location-stable-incubator-charts/


billimek (approved):
Thanks for fixing this!

PR #105, opened 2020-10-30
[unifi] syslog & promtail logging

michaelkoetter:
This adds two features:
- `syslogService` to capture syslog from Unifi devices (needs to be enabled in site configuration). Logs from Unifi devices will end up in `/unifi/logs/remote`.
- optional Promtail sidecar for controller logs and remote device logs if available

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/unifi/values.yaml
+++ b/charts/unifi/values.yaml
@@ -10,10 +10,10 @@ image:
   tag: 5.12.35
   pullPolicy: IfNotPresent
 
-# If enabled, the controller, discovery, GUI, and STUN services will not be
+# If enabled, the controller, discovery, GUI, STUN and syslog services will not be
 # created.
 # Instead, one service will be created with the port and nodePort settings from
-# controllerService, discoveryService, guiService, and stunService.
+# controllerService, discoveryService, guiService, stunService and syslogService.
 # This is useful if, for example, the ClusterIP network is routable and being
 # accessed directly by access points, and the APs don't have a way to discern
 # different services on different IPs.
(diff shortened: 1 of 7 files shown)


charts/unifi/values.yaml
```diff
@@ -169,6 +169,27 @@ discoveryService:
   ## Set the externalTrafficPolicy in the Service to either Cluster or Local
   # externalTrafficPolicy: Cluster
 
+syslogService:
+  type: NodePort
```
billimek, line 173: You used NodePort here because it needs to be reachable by UniFi devices on the network and thus ClusterIP wouldn't be a good choice, correct?

michaelkoetter: Yes, it should be configured the same way as `controllerService` or `stunService` and also respond to the same IP (ie. what devices consider the "controller address") to make it work, I think.

michaelkoetter: FWIW, I configured a `LoadBalancer` for all of the services which has the IP `192.168.30.158`, and then added a DNS entry `unifi` so that access points would automatically connect to that IP and register with the controller (what Unifi calls [Layer 3 Adoption](https://help.ui.com/hc/en-us/articles/204909754-UniFi-Layer-3-Adoption-for-Remote-UniFi-Controllers#6)).

After enabling the "remote syslog" option, access points got this configuration:

```bash
BedroomAP-BZ.v4.3.21# cat /tmp/system.cfg | grep syslog
# syslog
syslog.level=7
syslog.remote.encrypt=enabled
syslog.remote.ip=192.168.30.158
syslog.remote.key=<redacted>
syslog.remote.port=5514
syslog.remote.status=enabled
syslog.status=enabled
```
My guess is that the controller just configures the `syslog.remote.ip` with the IP that was determined as the "controller IP" during adoption, in my case because of the DNS entry - but there could be some other magic at play. Unfortunately, I couldn't find any official documentation on the Unifi site for this feature.

billimek (approved):
LGTM

PR #108, opened 2020-11-03
[home-assistant] Fix ports for home assistant

eddiezane:
This PR fixes reconfigured ports inside the home-assistant chart.

A `containerPort` should match the port of the application inside the container and in this case is not actually configurable. The `service.targetPort` must match the `deployment.....containerPort`. A `service.port` is the only thing we want to configure in this situation.

I ran into this when attempting to deploy a service of `LoadBalancer` on port 80.

An argument could be made that the vscode changes don't really fit here but I don't think the current implementation was ever working as the `codercom/code-server` image appears to listen on `8080` as opposed to the current default of `80`.

Not sure if this is just a patch version increase but as stated these things currently don't work.

#### Checklist

- [x] Chart Version bumped


charts/home-assistant/README.md
```diff
@@ -57,8 +57,8 @@ The following tables lists the configurable parameters of the Home Assistant cha
...
-| `service.portName`                              | Kubernetes port name where the home-assistant GUI is exposed                                                                                                                                                                              | `api`                                  |
+| `service.port`                                  | Port to be exposed by the home-assistant GUI service                                                                                                                                                                                      | `8123`                                 |
+| `service.portName`                              | Name of the port exposed within the home-assistant GUI service                                                                                                                                                                            | `api`                                  |
```
eddiezane, line 61: This field is unnecessary and could be removed but would likely be a breaking change?

onedr0p: Yes, in this case you'd want to bump the version to 3.0.0

onedr0p: If you are removing the config for ports, you can remove them from the readme too.

onedr0p: and make a note in the readme on the breaking changes, e.g. https://github.com/k8s-at-home/charts/tree/master/charts/blocky#upgrading-an-existing-release-to-a-new-major-version

eddiezane: @onedr0p would you be ok with shipping this as is and I'll open an issue to track removing it? Reason being I'm planning to dig in deep this weekend on Home Assistant and may find some other potentially breaking changes.

onedr0p: I'm a bit confused, you removed the variables in the templates but kept them in the readme?

eddiezane: Ah sorry for not being clear. I didn't remove the `service.portName` variable. I only cleaned up the description here. I think there may have been a misunderstanding when this chart was created that this port name needed to match the name in the deployment.

onedr0p: I think this is fine to merge, you are right that these apps don't have a current way to support changing the ports the apps are running on within the container.

onedr0p: or if they do, we don't give a way for people to change it.

charts/home-assistant/templates/deployment.yaml
```diff
@@ -332,7 +332,7 @@ spec:
           imagePullPolicy: {{ .Values.appdaemon.image.pullPolicy }}
           ports:
             - name: appdaemon
-              containerPort: {{ .Values.appdaemon.service.port }}
+              containerPort: 5050
               protocol: TCP
           env:
             - name: HA_URL
```
onedr0p, line 338: You can hardcode the line below this to `value: "http://localhost:8123"` ?

onedr0p (changes requested):

charts/home-assistant/templates/deployment.yaml
onedr0p: There's actually other places in this file that are referencing those ports too e.g.

```
            - name: HA_URL
              value: "http://localhost:{{ .Values.service.port }}"
            {{- if .Values.appdaemon.ingress.enabled }}
            - name: DASH_URL
              #value: http{{ if .Values.appdaemon.ingress.tls }}s{{ end }}://{{ index .Values.appdaemon.ingress.hosts 0 }}
              value: http://0.0.0.0:{{ .Values.appdaemon.service.port }}
```

eddiezane: Good catch. That will need to be `http://serviceName:8123`. I'll clean that up and check the others.

charts/home-assistant/templates/deployment.yaml
```diff
@@ -336,7 +331,7 @@ spec:
...
-              value: "http://localhost:{{ .Values.service.port }}"
+              value: "http://localhost:8123"
             {{- if .Values.appdaemon.ingress.enabled }}
             - name: DASH_URL
```
eddiezane, line 336: This one might need to stay as the bind address otherwise it won't listen correctly. I'll have to test https://github.com/AppDaemon/appdaemon/blob/3e141b15b28de9c60fe6e9481581d2748e620cae/dockerStart.sh#L37

onedr0p:
This should be rebased as there were other changes to home-assistant

onedr0p:
This PR is really out of sync with all the changes to the Hass chart recently, I am going to go ahead and close it.

We're planning on migrating this chart to use our common library, so this may not be needed but feel free to check back when that updated.

PR #147, opened 2020-11-13
[node-red] Add startupProbe

coldfire84:
This PR adds a startup probe to address an issue I ran into with pod startup taking longer than allowed liveness probe failures.

Approach outlined [here](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-startup-probes).

#### Checklist
- [x] Chart Version bumped


charts/node-red/templates/deployment.yaml
```diff
@@ -52,6 +52,12 @@ spec:
             httpGet:
               path: {{ .Values.readinessProbePath }}
               port: http
+          startupProbe:
```
billimek, line 55: Can we make this conditional?

startupProbe wasn't introduced until k8s 1.16 and will break those folks running prior to that version. 

Making it conditional will help with this possibility.

charts/node-red/templates/deployment.yaml
```diff
@@ -52,6 +52,12 @@ spec:
...
+            httpGet:
+              path: {{ .Values.startupProbePath }}
+              port: http
+            failureThreshold: 30
```
billimek, line 59: We should parameterize this so it can be adjusted.

charts/node-red/templates/deployment.yaml
```diff
@@ -52,6 +52,12 @@ spec:
...
+            periodSeconds: 10
```
billimek, line 60: We should parameterize this so it can be adjusted.

billimek (changes requested):
Just a few changes if you can

billimek (approved):
LGTM

Thank you for adding this great feature!

PR #149, opened 2020-11-13
[unifi] Add liveness and readiness probe configuration

tingvarsson:
* Parameterize timing of livenessProbe

* Parameterize timing of readinessProbe

* Bump version, no change in default behaviour

* Update README.md with new configuration parameters

Reason I want/need this configurable is a slow server stacked with other services as well. And I want to avoid using Resources to guarantee CPU time as it restrict how much I can cramp onto the server and the load is very sporadic from other services :smile: 
(note there might be smarter ways of solving my specific issue, but this feels like a low hanging fruit solution that I already utilize with other services as they provide this)

#### Checklist
- [x] Chart Version bumped


billimek (approved):
LGTM

PR #151, opened 2020-11-13
[zwave2mqtt] convert to use common library

onedr0p:

Major changes with the common library but totally wurf

#### Checklist
- [x] Chart Version bumped


charts/zwave2mqtt/values.yaml
```diff
@@ -1,105 +1,57 @@
...
-  ##
-  ## If you want to reuse an existing claim, you can pass the name of the PVC using
-  ## the existingClaim variable
-  # existingClaim: your-claim
-  accessMode: ReadWriteOnce
-  size: 1Gi
-  ## Do not delete the pvc upon helm uninstall
-  skipuninstall: false
+env:
+  # TZ
```
bjw-s, line 12: Maybe leave this out, since a default (example) is already set in `common`?

charts/zwave2mqtt/values.yaml
```diff
@@ -1,105 +1,57 @@
...
-
-affinity: {}
-
-podAnnotations: {}
+persistence:
+  config:
+    enabled: false
+    emptyDir: false
+    mountPath: /usr/src/app/store
+    ## Persistent Volume Storage Class
```
bjw-s, line 27: Do you want to keep these comments in? I don't think we have for the the other charts

charts/zwave2mqtt/values.yaml
```diff
@@ -1,105 +1,57 @@
...
+    # storageClass: "-"
+    # accessMode: ReadWriteOnce
+    # size: 1Gi
+    ## Do not delete the pvc upon helm uninstall
+    # skipuninstall: false
+    # existingClaim: ""
+
+additionalVolumeMounts:
+  - name: usb
+    mountPath: /dev/serial/by-id/usb-0658_0200-if00
```
bjw-s, line 42: Do you think we should put in a comment somewhere stating that this path can differ for the user? Maybe put it in the README as well?

bjw-s (commented):
Some small questions, but in general it looks great!

charts/zwave2mqtt/.helmignore
```diff
@@ -19,6 +19,5 @@
 .project
 .idea/
 *.tmproj
-.vscode/
```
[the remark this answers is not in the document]
onedr0p: I copied it from the other charts that were converted over. I agree it should be in here for all of them.

charts/zwave2mqtt/values.yaml
onedr0p: There's a default environment variable that needs to be set. or are you just talking about the TZ line?

charts/zwave2mqtt/values.yaml
onedr0p: For now i think it's fine to add in. 

we're really lacking on documentation so once we get that covered in a centralized wiki or something a lot of charts will be updated to reference it instead

charts/zwave2mqtt/values.yaml
bjw-s: Yes, I was only talking about the `# TZ` line 🙂

charts/zwave2mqtt/values.yaml
bjw-s: Sounds like a plan! 👍

charts/zwave2mqtt/values.yaml
onedr0p: Sure. Should that whole section be commented out? I'm not sure if helm will replace with the user values or merge with these.

bjw-s: No I just meant putting in a descriptive comment above the mountPath, stating something along the lines that this path should be double-checked by the user to make sure that they have a device on this path

bjw-s (approved):

bjw-s:
LGTM! 👍

PR #153, opened 2020-11-14
[alertmanager-bot] new chart

runningman84:
#### Checklist
- [x] Chart Version bumped
- [ ] Title of the PR starts with chart name (e.g. `[radarr]`)


runningman84:
@onedr0p are these tests supported to run the real workload? In this case we would need working telegram credentials which I do not want to store in the public....

onedr0p:
@runningman84 I am not sure, but you should be able to having the app run without specifying actual data... 

Since this is a new chart, do you think it might be a good idea to use the common library @bjw-s built?

https://github.com/k8s-at-home/charts/tree/master/charts/common

Here's an example of it being used, but there are 8 or so charts using it too.

https://github.com/k8s-at-home/charts/tree/master/charts/qbittorrent

Moving forward we'd like to see new charts use this if it isn't much of a hassle, it reduces maintenance and LoC across the charts.

onedr0p:
closing in favor of https://github.com/k8s-at-home/charts/pull/214

PR #159, opened 2020-11-15
[Plex] Added the ability to specify subPath on each extraData mount

gandazgul:
Added the ability to specify subPath on each of the extraData mounts

#### Checklist
- [x] Chart Version bumped
- [x] Variables are documented in the README.md - added to values.yaml


gandazgul:
@billimek @bjw-s Back with another subPath PR :) let me know if I missed anything. Thank you!

gandazgul:
Actually, I dont know what the difference is between extraMounts and extraData. What I really need is extraMounts but I think extraMounts shouldn't create a new volume, it should be as the name implies extraMounts from the same volumes.

My use case is I need to mount 2 different paths from the same claim, one has the movies and tv etc and the other has my pictures/home videos. With extra data it doesnt work because theres 2 volumes declared with the same claim. With extraMounts the same thing happens.

Would it be ok to change extraMounts to not create a volume? this would be breaking change, but I think is correct. Perhaps then I need to add a third extra* in persistence?

bjw-s:
I actually would like to see if the Plex chart can be migrated to the new common library, if there is going to be a new major version. That would provide all the required flexibility, but would also keep all/most of our charts working the same way.

onedr0p:
I agree with @bjw-s it would be nice to convert over the common library, but there's some advanced configuration happening in the deployment like the `plexPreferences:` and `certificate:` objects.

gandazgul:
I'm not sure if I know enough about the common library but I'd be willing to give it a go. Are there some docs and an example of a chart that is using the library?

onedr0p:
If you pull the latest master branch and run `grep -r "name: common" .` you'll see all the charts that use it. The comoon library is at https://github.com/k8s-at-home/charts/tree/master/charts/common

gandazgul:
@onedr0p thank you I'll take a look. As far as the change I want to make. 

Are you all ok with changing extraMounts to not create volumes but only generate mounts?

onedr0p:
This change is alright with me! It shouldn't break any existing setups.

onedr0p (approved):
👍

PR #168, opened 2020-11-16
[Plex] - Allow claimName to not be specified for extra mounts and add subPath

gandazgul:
My previous PR (https://github.com/k8s-at-home/charts/pull/159) didn't fully address my use case.

The use case here is I want to add another mount for the same volume data. This is somewhat unusual but I have 2 different paths in my data volume that I want plex to access; my movies/tv and my pictures.

This change will allow me to specify the pictures mount via extraMounts but attached to the same data volume.

#### Checklist
- [x] Chart Version bumped


gandazgul:
@bjw-s @onedr0p this pr is similar to my last but it fully addresses my use case. As you can see I found a way to do it without causing a breaking change.

onedr0p:
I'm a little confused. Why don't you mount the data as one volume into the container and tell plex where to find each directory (/nas/folder/tv /nas/folder/movies /nas/folder/photo) on that one volume?

gandazgul:
@onedr0p the volume has other things as well, I dont want to risk the chance of plex having access to them. I recognize it is a little unorthodox, but since I found a way to do it without disrupting more conventional setups I thought it would be ok to include this. 

thank you again for your time reviewing my PRs.

onedr0p:
Makes sense.. in a way, but I think having a folder that only Plex needs in your volume is a bit better. On my NAS I have a `/library/{tv,movies,music,photos}` folders and expose `/library` for Plex.

gandazgul:
@onedr0p yes, I agree. I keep my pictures in a different place. It has to do with the way I backup things. I backup my pictures along with my home folder and documents to an external server vs the movies, tv, etc only get backed up in a clone in the same server. Additionally other users with ssh access to the host get access to the public folder where TV and movies are but dont get access to my home folder and pictures. 

Hopefully this makes more sense as to why they are different locations.

onedr0p (approved):

gandazgul:
Thank you! @onedr0p

PR #170, opened 2020-11-16
[common] Read and the service ports correctly

### LINKED ISSUE #169: [common] Container fails to deploy when `service.enabled: false`
Common library has a `service.enabled` flag, but the container that is being generated doesn't check for that flag when setting the ports causing it to fail when applying.

_Originally posted by @bjw-s in https://github.com/k8s-at-home/charts/issues/167#issuecomment-727781761_

bjw-s:
Should fix #169

One edge case remains but I don't think we should build for it: if multiple ports would have the same name, the container spec will contain duplicate port names, which is not allowed.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/common/Chart.yaml
+++ b/charts/common/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: common
 description: Function library for k8s-at-home charts
 type: library
-version: 1.5.0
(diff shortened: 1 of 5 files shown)


onedr0p (approved):
👍

PR #175, opened 2020-11-17
[unifi] Add jvmInitHeapSize and jvmMaxHeapSize configuration

tingvarsson:
* Add jvmInitHeapSize and jvmMaxHeapSize which are passed on
  to the container as env vars

* Bump version, no change in default behaviour

* Update README.md with new configuration parameters

Signed-off-by: Thomas Ingvarsson <ingvarsson.thomas@gmail.com>

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/unifi/Chart.yaml
+++ b/charts/unifi/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 5.14.23
 description: Ubiquiti Network's Unifi Controller
 name: unifi
-version: 1.2.0
+version: 1.3.0
 keywords:
   - ubiquiti
   - unifi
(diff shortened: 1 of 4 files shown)


onedr0p:
@tingvarsson thanks for the contribution! Do you think those options could be set with the existing `extraJvmOpts` option?

tingvarsson:
@onedr0p looked some at it before the PR, jvmInitHeapSize (or rather -xms) is not set by upstream docker image by default and could be set via extraJvmOpts.
But jvmMaxHeapSize (-xmx) is set explicitly to a default value by upstream docker image, so setting it via extraJvmOpts would be overridden (not explicitly but by the ordering of the ARG string built up by the entrypoint script).

But a thought with the PR as is was to mimic the available config/env params as upstream docker image in this case.

onedr0p:
Thanks for the explanation, LGTM!

onedr0p (approved):

PR #180, opened 2020-11-18
[node-red] migrate to common library

onedr0p:
#### Checklist
- [x] Chart Version bumped


charts/node-red/README.md
```diff
@@ -1,107 +1,77 @@
-# Node-RED
+# Node-Red
 
-Low-code programming for event-driven applications
-
-**This chart is not maintained by the Node-RED project and any issues with the chart should be raised [here](https://github.com/k8s-at-home/charts/issues/new)**
```
bjw-s, line 5: This was actually requested by Node-RED here: https://discourse.nodered.org/t/node-red-helm-chart/3784/3 (even though it's been 2 years now...)

charts/node-red/values.yaml
```diff
@@ -1,132 +1,28 @@
...
-# Declare variables to be passed into your templates.
-
-# upgrade strategy type (e.g. Recreate or RollingUpdate)
-strategyType: Recreate
 
 image:
   repository: nodered/node-red
-  tag: 1.0.6-12-minimal
   pullPolicy: IfNotPresent
+  tag: 1.2.5
```
bjw-s, line 6: I don't know what the reasoning used to be to go for the `minimal` variant. But it also makes sense to me to go for the 'default' image by default and if people want to use another variant they can override the value

onedr0p: agreed!

bjw-s (approved):
LGTM 👍

PR #182, opened 2020-11-18
[zigbee2mqtt] set ZIGBEE2MQTT_DATA envar

onedr0p:
#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/zigbee2mqtt/values.yaml
+++ b/charts/zigbee2mqtt/values.yaml
@@ -8,8 +8,8 @@ image:
 strategy:
   type: Recreate
 
-env: {}
-  # TZ
+env:
+  ZIGBEE2MQTT_DATA: /data
 
 service:
   port:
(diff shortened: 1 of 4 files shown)


charts/zigbee2mqtt/values.yaml
```diff
@@ -8,8 +8,8 @@ image:
 strategy:
   type: Recreate
 
-env: {}
-  # TZ
+env:
+  ZIGBEE2MQTT_DATA: /data
```
bjw-s, line 12: It's good to have this default to gain persistence, but it can't start now since the /data folder is not enabled by default. 

Maybe set the defaults for `persistence.data.enabled` and `persistence.data.emptyDir`to `true`?

onedr0p: Maybe we can add it in the ct.yaml as an override for the ci while keeping it disabled in these values?

bjw-s (approved):
LGTM 👍

PR #190, opened 2020-11-23
add ISSUE_TEMPLATE

onedr0p:

Create issue templates for the charts repo

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -1,0 +1,24 @@
+---
+name: Bug report
+about: Create a report to help us improve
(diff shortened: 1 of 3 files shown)


carpenike:
Is it worthwhile to pre-embed the title with the request type?

bjw-s:
> Is it worthwhile to pre-embed the title with the request type?

Isn't that also covered by the labels? Or do you think it would have benefit to add "[<chart-name>] " as a prefix to the bug/enhancement titles?

carpenike:
Labels should be fine as well as they show up in the default table view.

onedr0p:
We can always update these later on with improvement, this is a good base

LGTM :)

PR #195, opened 2020-11-23
[workflow] update linting and release workflows

onedr0p:

- moved `ct.yaml` to `.github` directory
- updated linting workflow and bumped `helm/chart-testing-action` to `v2.0.1`
- updated release workflow and bumped `helm/chart-releaser-action` to `v1.1.0`
- added job after release job to update `generated` timestamp in `index.yaml` in the `gh-pages` branch

Minor issue with `helm/chart-releaser-action` is this issue https://github.com/helm/chart-releaser-action/issues/61 however I didn't see this behavior in our tests

### DIFF
--- a/.github/workflows/lint-test.yaml
+++ b/.github/workflows/lint-test.yaml
@@ -30,9 +30,11 @@ jobs:
           if [[ -n "$changed" ]]; then
             echo "::set-output name=changed::true"
           fi
+
       - name: Run chart-testing (lint)
         id: lint
         run: ct lint --config .github/ct.yaml
+        if: steps.list-changed.outputs.changed == 'true'
 
       - name: Create kind cluster
         uses: helm/kind-action@v1.1.0
@@ -40,3 +42,4 @@ jobs:
 
       - name: Run chart-testing (install)
         run: ct install --config .github/ct.yaml
+        if: steps.list-changed.outputs.changed == 'true'
--- a/.github/workflows/lint-test.yaml
+++ b/.github/workflows/lint-test.yaml
@@ -1,28 +1,42 @@
 name: Lint and Test Charts
+
 on: pull_request
+
 jobs:
   lint-test:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-      - name: Fetch history
+        with:
+          fetch-depth: 0
+
+      - name: Install Helm
+        uses: azure/setup-helm@v1
+        with:
+          version: v3.4.0
+
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
+
+      - name: Set up chart-testing
+        uses: helm/chart-testing-action@v2.0.1
+
+      - name: Run chart-testing (list-changed)
+        id: list-changed
         run: |
-          git fetch --prune --unshallow;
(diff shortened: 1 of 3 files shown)


onedr0p:
Let's get a couple rounds of review to make sure I didn't fat finger anything :D

.github/workflows/lint-test.yaml
```diff
@@ -1,28 +1,42 @@
...
+        uses: helm/kind-action@v1.1.0
+        if: steps.list-changed.outputs.changed == 'true'
+
       - name: Run chart-testing (install)
-        uses: helm/chart-testing-action@v1.0.0
-        if: "steps.lint.outputs.changed == 'true' && ! contains(env.commitmsg, '[skip install]')"
-        with:
-          command: install
-          config: ct.yaml
+        run: ct install --config .github/ct.yaml
```
bjw-s, line 42: I'm wondering if we should add the `if: steps.list-changed.outputs.changed == 'true'` to both the `ct lint` and and `ct install` steps as well. It won't be a huge time saver, but it prevents unnecessary steps.

.github/workflows/release.yaml
```diff
@@ -21,25 +24,41 @@ jobs:
...
+          ref: "gh-pages"
+          fetch-depth: 0
+
+      - name: Commit and push timestamp updates
+        run: |
+          git config --global user.email "$GITHUB_ACTOR"
+          git config --global user.name "$GITHUB_ACTOR@users.noreply.github.com"
+          export generated_date=$(date --utc +%FT%T.%9NZ)
+          sed -i -e "s/^generated:.*/generated: \"$generated_date\"/" index.yaml
+          git add index.yaml
```
bjw-s, line 62: This will fail if there is no `index.yaml` present in the repo. Not an issue anymore on this repo, but maybe for somebody who wants to use this setup as inspiration for for their own new repo we could make it more robust by adding an `if [[ -f index.yaml ]]; `

.github/workflows/release.yaml
```diff
@@ -21,25 +24,41 @@ jobs:
...
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          ref: "gh-pages"
+          fetch-depth: 0
+
+      - name: Commit and push timestamp updates
+        run: |
+          git config --global user.email "$GITHUB_ACTOR"
+          git config --global user.name "$GITHUB_ACTOR@users.noreply.github.com"
```
bjw-s, line 59: Technically not required, but maybe put the git config commands in a separate step like in the `release` job?

onedr0p: In our tests with in the other repo git failed because these were not set. I'll update to put it in its own job here

bjw-s: Oh I see my comment was a bit ambiguous... What I meant was that it's technically not required to move it to another step 😅

.github/workflows/lint-test.yaml
onedr0p: updated!

onedr0p:
All fixed up :)

billimek (approved):
LGTM

onedr0p:
Shipping!!

PR #78, opened 2020-10-02
[powerdns] flip default to postgres, fix postgres dependent chart

carpenike:
#### Special notes for your reviewer:
Found a bunch of weird stuff that I overlooked...

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [X] Chart Version bumped
- [X] Variables are documented in the README.md
- [X] Title of the PR starts with chart name (e.g. `[radarr]`)

PR #215, opened 2020-11-25
[home-assistant] update postgresql to use right options

onedr0p:

Bumping major...

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/home-assistant/CHANGELOG.md
+++ b/charts/home-assistant/CHANGELOG.md
@@ -4,11 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [4.0.0]
+## [3.1.0]
 
 ### Changed
 
 - Fixed issue in default postgresql configuration.
+- Bumped bitnami/postgresql to 10.1.1 in chart deps
 
 ## [3.0.0]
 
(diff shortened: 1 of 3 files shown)


carpenike:
Let's make sure #78 works first.

onedr0p:
I think #78 is a runtime issue, in hass we don't enable the databases by default

charts/home-assistant/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 0.118.3
 description: Home Assistant
 name: home-assistant
-version: 3.0.0
+version: 4.0.0
```
bjw-s, line 5: Why did you make this a major version bump? It's a bugfix on 3.0.0 if you ask me... 

The values should have already been in the values.yaml in 3.0.0, since the postgresql chart requires them there. If we would have put the values in the correct spot in our own values files it would have worked, we are only providing the correct defaults now.

charts/home-assistant/CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [4.0.0]
```
bjw-s, line 7: It's probably bad practice to retroactively modify changelog/release notes for a previous version, but maybe we should add something like this to the `postgresql` upgrade in 3.0.0?

`Any pre-existing StatefulSet will have to be removed before upgrading due to a name change in the chart.`

bjw-s (changes requested):

bjw-s:
Maybe pick up #217 in one go?

bjw-s (approved):
LGTM

PR #252, opened 2020-11-26
[paperless] New chart

mr-onion-2:
My first contribution to this great project.

This relies on a small tweak to the common chart (#253)

Hope this is all OK. Happy to make any changes you suggest.

Thanks

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [X] Chart Version bumped
- [X] Variables are documented in the README.md
- [X] Title of the PR starts with chart name (e.g. `[radarr]`)

PR #253, opened 2020-11-26
[common] Add args option to mainContainer

mr-onion-2:
Required for #252 

#### Checklist
- [X] Chart Version bumped
- [X] Variables are documented in the README.md
- [X] Title of the PR starts with chart name (e.g. `[radarr]`)


charts/common/templates/lib/controller/_container.tpl
```diff
@@ -5,6 +5,9 @@ The main container included in the controller.
 - name: {{ include "common.names.fullname" . }}
   image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
   imagePullPolicy: {{ .Values.image.pullPolicy }}
+  {{- if .Values.image.args }}
```
onedr0p, line 8: I'm thinking this should be .Values.args

what do you think at @bjw-s ?

onedr0p (changes requested):
Thanks for the change. 

You should also put this new field in the values.yaml. 

```yaml
args: []
```

charts/common/templates/lib/controller/_container.tpl
bjw-s: Agreed 👍

charts/common/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: common
 description: Function library for k8s-at-home charts
 type: library
-version: 1.6.2
+version: 1.6.3
```
bjw-s, line 5: New functionality is being added here so this should be `1.7.0`

bjw-s (changes requested):

charts/common/templates/lib/controller/_container.tpl
mr-onion-2: Good call. I did ponder on that. Fixed!

charts/common/values.yaml
```diff
@@ -12,6 +12,9 @@ strategy:
   ## DaemonSets ignore this
   type: RollingUpdate
 
+# Override the default command for the mainContainer image
```
onedr0p, line 15: This can simply say

`Override the default args`

command and args are different things

onedr0p (changes requested):
one minor update on a comment and i think we're good!

charts/common/values.yaml
mr-onion-2: Thanks

charts/common/templates/lib/controller/_container.tpl
```diff
@@ -5,8 +5,8 @@ The main container included in the controller.
 - name: {{ include "common.names.fullname" . }}
   image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
   imagePullPolicy: {{ .Values.image.pullPolicy }}
-  {{- if .Values.image.args }}
-  args: {{ .Values.image.args }}
+  {{- if .Values.args }}
+  args: {{ .Values.args }}
```
bjw-s, line 9:
```suggestion
  {{- with .Values.args }}
  args: {{ . }}
```

This would be even cleaner. Apoligies, I missed that in the first review

bjw-s (approved):

onedr0p (approved):

Issue #10, opened 2020-06-14
[home-assistant] Homeassistant fails to start with AppDaemon enabled


rwlove:
**Describe the bug**
I'm instantiating HASS in a kubernetes cluster using the following definition. https://github.com/billimek/billimek-charts/tree/master/charts/home-assistant
 
Homeassistant fails to start with AppDaemon enabled.

**A clear and concise description of what the bug is.**
Version of Helm and Kubernetes:
Helm: v3.2.1
Kubernetes: v1.18.2

What happened:
```
➜  kubernetes git:(master) ✗ kubectl -n homeassistant logs homeassistant-home-assistant-8ccc9d79-f22bk -c appdaemon 
2020-06-14 16:35:45.284536 INFO AppDaemon Version 3.0.5 starting
2020-06-14 16:35:45.284770 INFO Configuration read from: /conf/appdaemon.yaml
2020-06-14 16:35:45.286422 INFO AppDaemon: Starting Apps
2020-06-14 16:35:45.289228 INFO AppDaemon: Loading Plugin HASS using class HassPlugin from module hassplugin
2020-06-14 16:35:45.329504 INFO AppDaemon: HASS: HASS Plugin Initializing
2020-06-14 16:35:45.329869 INFO AppDaemon: HASS: HASS Plugin initialization complete
2020-06-14 16:35:45.330040 INFO Starting Dashboards
2020-06-14 16:35:45.330182 WARNING ------------------------------------------------------------
2020-06-14 16:35:45.330251 WARNING Unexpected error during run()
2020-06-14 16:35:45.330311 WARNING ------------------------------------------------------------
2020-06-14 16:35:45.330992 WARNING Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/appdaemon/admain.py", line 82, in run
    self.rundash = rundash.RunDash(self.AD, loop, self.logger, self.access, **hadashboard)
  File "/usr/local/lib/python3.6/site-packages/appdaemon/rundash.py", line 130, in __init__
    dash_net = url.netloc.split(":")
TypeError: a bytes-like object is required, not 'str'

2020-06-14 16:35:45.331065 WARNING ------------------------------------------------------------
2020-06-14 16:35:45.331140 INFO AppDeamon Exited
```

Relevant Configuration Section
```
appdaemon:
  enabled: true

  ## code-server container image                                                                                                              
  ##                                                                                                                                          
  image:
    repository: acockburn/appdaemon
[...]

PR #305, opened 2020-12-06
[home-assistant] Fix env for appdaemon

lexfrei:

Looks like there was a copy-paste error and vscode extra vars was passed to appdaemon. Also:

- More flexible env configuration
- Now we can use secrets for token 

This change replaces `extraEnv` with `env` and `envFrom` and changes syntax. This doesn't look braking just coz it never worked before, but it's not consistent with else containers in this deployment. I can't imagine how should I change this to be consistent and get the same flexibility. If breaking changes is fine, I can replace all else envs with this syntax.

#### Checklist

- [x] Bump chart version

#### Related:
#10


lexfrei:
Also, I think we should remove all preset envs and just mention them in a comment before env block. But it's breaking changes so not done yet

onedr0p (approved):

onedr0p:
Thanks for the PR @lexfrei 👍

lexfrei:
@onedr0p should i rewrite env templateing for whole deployment?
They will be more useful, but the change will break backward compatibility.

onedr0p:
Sure! Let's see what you're thinking of in the PR just make sure to bump major version and write in the changelog

PR #326, opened 2020-12-08
[appdaemon] create separate chart outside of home-assistant

onedr0p:

Step 1 of removing from the home-assistant chart

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/appdaemon/ci/ct-values.yaml
+++ b/charts/appdaemon/ci/ct-values.yaml
@@ -1,0 +1,2 @@
+ingress:
+  enabled: false
--- a/charts/appdaemon/ci/ct-values.yaml
+++ b/charts/appdaemon/ci/ct-values.yaml
@@ -5,4 +5,4 @@ env:
   ELEVATION: "64"
   DASH_URL: "http://0.0.0.0:5050"
 ingress:
-  enabled: false
+  enabled: true
--- a/charts/appdaemon/ci/ct-values.yaml
+++ b/charts/appdaemon/ci/ct-values.yaml
@@ -1,7 +1,7 @@
 env:
   TZ: "Africa/Cairo"
-  LAT: "29.9792"
-  LONG: "31.1342"
+  LATITUDE: "29.9792"
+  LONGITUDE: "31.1342"
   ELEVATION: "64"
   DASH_URL: "http://0.0.0.0:5050"
 ingress:
--- a/charts/appdaemon/ci/ct-values.yaml
+++ b/charts/appdaemon/ci/ct-values.yaml
@@ -1,7 +1,2 @@
-env:
-  TZ: "Africa/Cairo"
-  LAT: "29.9792"
-  LONG: "31.1342"
-  ELEVATION: "64"
 ingress:
   enabled: false
--- a/charts/appdaemon/ci/ct-values.yaml
+++ b/charts/appdaemon/ci/ct-values.yaml
@@ -1,2 +1,8 @@
+env:
+  TZ: "Africa/Cairo"
+  LAT: "29.9792"
+  LONG: "31.1342"
+  ELEVATION: "64"
+  DASH_URL: "http://0.0.0.0:5050"
 ingress:
   enabled: false
--- a/charts/appdaemon/ci/ct-values.yaml
+++ b/charts/appdaemon/ci/ct-values.yaml
@@ -1,2 +1,6 @@
+env:
+  TZ: "Africa/Cairo"
+  LAT: "29.9792"
+  LONG: "31.1342"
 ingress:
   enabled: false
--- a/charts/appdaemon/ci/ct-values.yaml
+++ b/charts/appdaemon/ci/ct-values.yaml
@@ -2,5 +2,6 @@ env:
   TZ: "Africa/Cairo"
   LAT: "29.9792"
   LONG: "31.1342"
+  ELEVATION: "64"
 ingress:
   enabled: false
--- a/charts/appdaemon/values.yaml
+++ b/charts/appdaemon/values.yaml
@@ -1,0 +1,86 @@
+# Default values for Appdaemon.
+
+image:
+  repository: acockburn/appdaemon
+  pullPolicy: IfNotPresent
+  tag: 4.0.5
+
+strategy:
+  type: Recreate
+
+service:
+  port:
+    port: 5050
+  # # Boilerplate config for code-server functionality, be sure to make changes to suit your environment
+  # additionalServices:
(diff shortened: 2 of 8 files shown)


charts/appdaemon/values.yaml
```diff
@@ -0,0 +1,86 @@
...
+  pullPolicy: IfNotPresent
+  tag: 4.0.5
+
+strategy:
+  type: Recreate
+
+service:
+  port:
+    port: 5050
+  # # Boilerplate config for code-server functionality, be sure to make changes to suit your environment
```
hyperbolic2346, line 14: Boilerplate config for vscode-server?

charts/appdaemon/values.yaml
```diff
@@ -0,0 +1,86 @@
...
+  #   type: ClusterIP
+  #   port:
+  #     port: 12321
+  #     name: vscode
+  #     protocol: TCP
+  #     targetPort: 12321
+
+env: {}
+  # HA_URL:
+  # DASH_URL:
```
hyperbolic2346, line 27: Should these be quoted in the values.yaml? Do you include http/https? Maybe add an example in here?

charts/appdaemon/values.yaml
hyperbolic2346: Is it called code-server? I'm not sure what is correct here. ;)

onedr0p: It is called code-server :)

https://github.com/cdr/code-server

charts/appdaemon/values.yaml
```diff
@@ -0,0 +1,86 @@
...
+    ## If defined, storageClassName: <storageClass>
+    ## If set to "-", storageClassName: "", which disables dynamic provisioning
+    ## If undefined (the default) or set to null, no storageClassName spec is
+    ##   set, choosing the default provisioner.  (gp2 on AWS, standard on
+    ##   GKE, AWS & OpenStack)
+    # storageClass: "-"
+    # accessMode: ReadWriteOnce
+    # size: 1Gi
+    ## Do not delete the pvc upon helm uninstall
+    # skipuninstall: false
```
hyperbolic2346, line 85: This is confusing to me. `skipuninstall: false` says to leave the pvc when uninstalled via helm?

charts/appdaemon/values.yaml
onedr0p: I'll add a link that refers to the documentation over at appdaemons docs

hyperbolic2346: I like that, but the general question is more of a helm formatting thing. It's always nice to see an example you can edit.
```
env:
  #  HA_URL: "http://home.mydomain.com:8123"
  #  DASH_URL: http://home.mydomain.com:5050
```
Those are two ways to do it that people could use. Would both quoted and non work? My point here is that if we fill in with even one example people can follow that and not have to guess or go to the helm docs/google.

onedr0p: I added a link to their docs, I don't think we need to include examples. We shouldn't need to add much documentation and have people refer to the upstream docs.

charts/appdaemon/values.yaml
onedr0p: This is carried over for all the charts using the common library 😆  It adds a helm hook to the PVC.

```metadata:
  name: {{ $pvcName }}
  {{- if or $values.skipuninstall $values.annotations }}
  annotations:
    {{- if $values.skipuninstall }}
    "helm.sh/resource-policy": keep
    {{- end }}```

So, if true, keep PVC when uninstalling with helm.

File: charts/appdaemon/values.yaml (hunk 2 above)
Comment (bjw-s) [reply charts/appdaemon/values.yaml:29]: I agree with @onedr0p . As a rule of thumb we try to avoid pre-filling container configuration if it can be avoided. 
This (hopefully 🙂) stimulates chart users to look at the upstream docs so they can gain an understanding of the settings they wish to modify, instead of changing an arbitrary value that we are providing as a default.

File: charts/appdaemon/values.yaml (hunk 3 above)
Comment (hyperbolic2346) [reply charts/appdaemon/values.yaml:87]: Right, the comment implies that setting it to false would keep the pvc, but that doesn't seem to match what I would expect, or what is actually happening here.

Comment (onedr0p) [reply charts/appdaemon/values.yaml:87]: Very true, I think we need to update this across all charts starting with this one. Good catch!

**Comment by onedr0p** [comment]
Hm.. looks like a `/conf/appdaemon.cfg` is required and needs at least the following config:

```
appdaemon:
  latitude: xxxx.xxxx
  longitude: -xxxx.xxxx
  elevation: xx
  time_zone: us/pacific
```

Bummer it doesn't pick it up from the environment variables :(

**Comment by onedr0p** [comment]
@alexdor will be building a custom image that hopefully takes care of this issue :)

**Comment by onedr0p** [comment]
I am having difficulties getting the custom image I built to work correctly 🙄 

Seems like an issue with Python on start :/

**Comment by hyperbolic2346** [comment]
maybe try an older python? I saw an issue saying it didn't work with 3.9, but maybe 3.8 as well?

**Comment by onedr0p** [comment]
I pinned to 3.8 but also had issues. I think we should revert to 3.0.5 until the following PR is merged 

https://github.com/AppDaemon/appdaemon/pull/1002

**Comment by onedr0p** [comment]
This _should_ be ready for review!

**Comment by onedr0p** [comment]
Finally ready for review @bjw-s this was a fun char tot work on 🙄

File: charts/appdaemon/ci/ct-values.yaml [hunk 4]
```diff
@@ -0,0 +1,8 @@
+env:
+  TZ: "Africa/Cairo"
+  LATITUDE: "29.9792"
+  LONGITUDE: "31.1342"
+  ELEVATION: "64"
+  DASH_URL: "http://0.0.0.0:5050"
+ingress:
```
bjw-s, line 7: This shouldn't be necessary as it's the default, right?

bjw-s (approved):
Looks good to me, small question on the ct-values, but I'm fine with it being left in too.

charts/appdaemon/ci/ct-values.yaml
onedr0p: appdaemon needs all those environment variables set or it will not pass CI :(

bjw-s: I meant the ingress, maybe I selected the wrong line 😊

onedr0p: Got it, I updated to enable ingress

PR #327, opened 2020-12-09
[home-assistant] deprecate appdaemon and configurator sidecars

onedr0p:
#### Special notes for your reviewer:

- Waiting on https://github.com/k8s-at-home/charts/pull/326
- VSCode seems like the popular choice over configurator

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [x] Chart Version bumped
- [x] Variables are documented in the README.md
- [x] Title of the PR starts with chart name (e.g. `[radarr]`)

PR #330, opened 2020-12-09
[homeassistant] deploy ha as statefulset instead of deployment

jffz:
This change might be useful for some ha integrations which register a token based on `socket.gethostname()` method like [freebox integration](https://github.com/home-assistant/core/blob/dev/homeassistant/components/freebox/const.py#L16).

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/home-assistant/Chart.yaml
+++ b/charts/home-assistant/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v2
-appVersion: 0.118.3
+appVersion: 0.118.5
 description: Home Assistant
 name: home-assistant
-version: 3.4.0
+version: 3.4.1
 keywords:
 - home-assistant
 - hass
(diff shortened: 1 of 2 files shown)


onedr0p:
@jffz I think we're looking to integrated this chart into using the common library like most of our other charts.

Once that is done using statuefulset will be an option to the end user.

jffz:
I am still discovering about the helm / charts. What is currently preventing ha to be using the common library ?

onedr0p:
These 2 PRs will help us get to the goal of using our common library

https://github.com/k8s-at-home/charts/pull/327
https://github.com/k8s-at-home/charts/pull/326

onedr0p:
It's also worth noting that `socket.gethostname()` will get the hostname of the container, not the host...

jffz:
> It's also worth noting that `socket.gethostname()` will get the hostname of the container, not the host...
> 
> Edit: or not! TIL each Pod in a StatefulSet derives its hostname from the name of the StatefulSet and the ordinal of the Pod.

That's the point of moving the deployment to a statefulset as any registered token using `socket.gethostname()` will be outdated on a pod deletion/recreation. Using a statefulset will always result in <statefulset-name>-<id> which will be homeassistant-0 in 99,99% of the cases.

onedr0p:
@jffz please follow this issue https://github.com/k8s-at-home/charts/issues/331

PR #346, opened 2020-12-12
[bitwardenrs] Update to 1.18.0 and add some missing features

alebastr:
* Add configuration for Yubico OTP server
* Add external database configuration
* Generate Ingress rules for websocket notification support.
  Not the best way to fix this, but I haven't figured out a better one without breaking ingress values format.

#### Checklist
- [x] Chart Version bumped
- [ ] Variables are documented in the README.md

### DIFF
--- a/charts/bitwardenrs/templates/configmap.yaml
+++ b/charts/bitwardenrs/templates/configmap.yaml
@@ -44,6 +44,7 @@ data:
   {{- end }}
   {{- end }}
   {{- end }}
+  {{- include "bitwardenrs.externalDatabaseConfigMap" . | nindent 2 }}
   {{- if .Values.env }}
   {{- toYaml .Values.env | nindent 2 }}
   {{- end }}
--- a/charts/bitwardenrs/templates/configmap.yaml
+++ b/charts/bitwardenrs/templates/configmap.yaml
@@ -12,8 +12,8 @@ data:
   WEBSOCKET_ENABLED: {{ .Values.bitwardenrs.websockets.enabled | quote }}
   {{- if and .Values.bitwardenrs.admin.enabled .Values.bitwardenrs.admin.disableAdminToken }}
   DISABLE_ADMIN_TOKEN: "true"
-    {{- end }}
-    {{- with .Values.bitwardenrs.smtp }}
+  {{- end }}
+  {{- with .Values.bitwardenrs.smtp }}
   {{- if .enabled }}
   SMTP_HOST: {{ required "SMTP host is required to enable SMTP" .host | quote }}
   SMTP_FROM: {{ required "SMTP sender address ('from') is required to enable SMTP" .from | quote }}
@@ -28,6 +28,19 @@ data:
   {{- end }}
   {{- if and (not .existingSecret.enabled) .user }}
   SMTP_USERNAME: {{ .user | quote }}
+  {{- end }}
+  {{- end }}
+  {{- end }}
+  {{- with .Values.bitwardenrs.yubico }}
+  {{- if .enabled }}
+  {{- if .server }}
+  YUBICO_SERVER: {{ .server | quote }}
+  {{- end }}
+  {{- if and (not .existingSecret.enabled) .clientId }}
+  YUBICO_CLIENT_ID: {{ .clientId | quote }}
+  {{- end }}
+  {{- if and (not .existingSecret.enabled) .secretKey }}
+  YUBICO_SECRET_KEY: {{ .secretKey | quote }}
   {{- end }}
   {{- end }}
   {{- end }}
--- a/charts/bitwardenrs/templates/deployment.yaml
+++ b/charts/bitwardenrs/templates/deployment.yaml
@@ -54,11 +54,11 @@ spec:
                   {{- else }}
                   name: {{ $fullName }}
                   key: admin-token
-                {{- end }}
-                {{- end }}
-                {{- end }}
-                {{- with .Values.bitwardenrs.smtp }}
-                {{- if eq .enabled true }}
+                  {{- end }}
+            {{- end }}
+            {{- end }}
+            {{- with .Values.bitwardenrs.smtp }}
+            {{- if eq .enabled true }}
             {{- if and .existingSecret.enabled (not .user) }}
             - name: SMTP_USERNAME
               valueFrom:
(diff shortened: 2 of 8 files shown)


onedr0p:
@DirtyCajunRice and @carpenike I know you guys use this chart, please review when you can a moment. Thanks!

charts/bitwardenrs/templates/deployment.yaml
```diff
@@ -70,9 +73,25 @@ spec:
...
+                secretKeyRef:
+                  name: {{ .existingSecret.name | quote }}
+                  key: {{ .existingSecret.clientIdKey | quote }}
+            {{- end }}
+            - name: YUBICO_SECRET_KEY
+              valueFrom:
+                secretKeyRef:
+                  name: {{ .existingSecret.name | quote }}
+                  key: {{ .existingSecret.secretKeyKey | quote }}
+            {{- end }}
```
DirtyCajunRice, line 93: the above would be related to the first comment

charts/bitwardenrs/templates/deployment.yaml
```diff
@@ -42,6 +42,9 @@ spec:
           envFrom:
             - configMapRef:
                 name: {{ $fullName }}
+            {{- with .Values.envFrom }}
+            {{- toYaml . | nindent 12 }}
+            {{- end }}
```
DirtyCajunRice, line 47: Instead of envFrom, which would require an all-inclusive addition that already exists, thoughts on the same format but just env? so that users can add individual additional k/v?

charts/bitwardenrs/templates/configmap.yaml
```diff
@@ -31,6 +31,19 @@ data:
...
+  {{- if .enabled }}
+  {{- if .server }}
+  YUBICO_SERVER: {{ .server | quote }}
+  {{- end }}
+  {{- if and (not (and .existingSecret.enabled .existingSecret.clientIdKey)) .clientId }}
+  YUBICO_CLIENT_ID: {{ .clientId | quote }}
+  {{- end }}
+  {{- if and (not .existingSecret.enabled) .secretKey }}
+  YUBICO_SECRET_KEY: {{ .secretKey | quote }}
+  {{- end }}
```
DirtyCajunRice, line 44: the above is fine, but what are your thoughts on all or none? if someone is going to put the key in a secret, chances are they will put the ID there as well.

DirtyCajunRice (commented):
Overall great, just a couple thoughts.

charts/bitwardenrs/templates/deployment.yaml
alebastr: The all-inclusive addition in this chart only supports configmap. As the comment in values.yaml suggests, I was looking for a way to read DATABASE_URL from a secret as it contains database password.

My first idea was
```yaml
envFromSecrets:
  DATABASE_URL:
    secret: bitwardenrs-db
    key: databaseUrl
```
but then I considered that the direction of this repository is to use the `common` template for charts or at least a common structure for values and decided to take the same approach.

And of course all of that is just an excuse to avoid adding proper database settings object :slightly_smiling_face:

charts/bitwardenrs/templates/configmap.yaml
alebastr: I don't have a strong opinion on that.
Requiring that both variables should be defined in a secret would simplify conditions though, so I'll do that.

charts/bitwardenrs/templates/deployment.yaml
alebastr: I added an external database config object so my configuration no longer requires `envFrom`.
It's still the only way to configure Duo 2FA or HIBP API key, so I'd prefer to have a support for reading arbitrary environment variables from secret. If you have any thoughts on the format for configuration, I can try implementing that.

charts/bitwardenrs/values.yaml
```diff
@@ -24,6 +24,31 @@ bitwardenrs:
...
       name: ""
       tokenKey: ""
+  # External database configuration.
+  # Requires bitwardenrs/server >= 1.17.0 or bitwardenrs/server-{mysql,postgres} images
+  # ref: https://github.com/dani-garcia/bitwarden_rs/wiki/Using-the-MySQL-Backend
+  #      https://github.com/dani-garcia/bitwarden_rs/wiki/Using-the-PostgreSQL-Backend
+  externalDatabase:
+    enabled: false
+    # Supported values: 'mysql', 'postgresql'.
+    type: ""
```
alebastr, line 34: It might be worth implementing this as a `database: "sqlite" | "mysql" | "postgresql"` and configuring details for the specific database backend in a subchart options format, i.e.
```yaml
bitwardenrs:
  database: "postgresql"

postgresql:
   postgresqlPassword: ... 
```
Let me know it's preferred to do things that way.

onedr0p: @alebastr that makes sense to me, but would be a breaking change.

carpenike: @onedr0p -- how about we introduce the breaking change with the default value of the old version? Bump major version and include reference to the breaking change in the readme.

alebastr: Oh, sorry, seems like I missed the notification with @onedr0p's response. 
I don't think it's breaking if the database support exists only within this PR and the default is an old behavior (sqlite). IMO, it's the right moment to make decision on that without breaking changes.

That being said, I still need to learn how to use dependency charts before implementing this change and I didn't have time to do that :smile:

onedr0p: > @onedr0p -- how about we introduce the breaking change with the default value of the old version? Bump major version and include reference to the breaking change in the readme.

this is fine with me as well

CuBiC3D:
What's the status here? 1.18.0 is released already, maybe a good point to think about the new changes like config file possible now.

carpenike:
Let's move to the new format which enables different db backends but set the default values to support legacy mode.

Update the readme in the breaking changes section and ensure the chart has a major version bump.

alebastr:
> Update the readme in the breaking changes section and ensure the chart has a major version bump.

Done.

charts/bitwardenrs/values.yaml
alebastr: My attempts to pull external database configuration from dependency chart config section ended in vain for a couple of reasons:
 - getting database host/port from postgresql or mariadb configuration requires uncomfortable amounts of internal knowledge about the structure of these charts
 - password value for bitwardenrs should be percent-encoded and therefore cannot be shared with other charts

Thus, the `bitwardenrs.externalDatabase` section remains as is unless someone else figures out a better way to do this.
It might be still worthy to declare these dependencies in the chart though.

carpenike (approved):
LGTM! Thanks!

PR #353, opened 2020-12-14
[blocky] Update to 0.11 and add query log pvc

ajc161:
**Description of the change**

Adds a volume to the blocky deployment for query logs. Can either be an `emptyDir` if persistence is disabled, an existing PVC referenced by `persistence.existingClaim`, or new PVC.

**Benefits**

Allows the use of query logs without filling the pod's ephemeral storage. Persistence is opt-in and will only be used if query logs are enabled.

**Possible drawbacks**

Only supports a deployment with a referenced PVC for now, using a StatefulSet is not supported. This means that only 1 replica can be used if the access mode is `ReadWriteOnce`. Additionally the PVC will be deleted if the chart is uninstalled.

**Checklist** 

### DIFF
--- a/charts/blocky/Chart.yaml
+++ b/charts/blocky/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v2
-appVersion: v0.10
+appVersion: v0.11
 description: DNS proxy as ad-blocker for local network
 name: blocky
-version: 4.0.1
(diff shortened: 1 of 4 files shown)


billimek (approved):
LGTM

Thanks for adding this feature!

Issue #137, opened 2020-11-10
[plex] - Plex fails to start because of readiness checks


yasn77:
When Plex is managing large libraries, it can take a really long time to start (so readiness checks fail and restart the pod). This is because it tries to chown everything inside of `/config`, and that directory might be big (in terms of the number of files) if you have a large library.

I've created a [PR in upstream plex Docker image](https://github.com/plexinc/pms-docker/pull/61) which should hopefully speed it up a bit.

I've also resorted to using `CHANGE_CONFIG_DIR_OWNERSHIP` set to `false` (this brought up a templating issue, for which I'll create a PR), which works but imho isn't ideal.

I was thinking the better approach might be to create an init container or possibly even a job that sets the ownership correctly... This way the readiness/liveness probes only are dealing with actual Plex starting up, not any preamble that might need to happen.

I would be interested in ideas.

bjw-s:
This seems like a case that could match the use case for startup probes: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-startup-probes

> Sometimes, you have to deal with legacy applications that might require an additional startup time on their first initialization. In such cases, it can be tricky to set up liveness probe parameters without compromising the fast response to deadlocks that motivated such a probe 

One downside could be that startupProbe is only available (by default) in K8s 1.18 onward.

billimek:
+1 to the idea for using StartupProbes.  I think the init container is also an interesting idea.  Maybe a combination of the two?

onedr0p:
With the release of kubernetes 1.20 our minimum support kubernetes version on these charts is 1.17. If people are on 1.17 or 1.16 they can enable the FeatureGate for StartUpProbes.

PR #361, opened 2020-12-16
[plex] Quote values for extraEnv

yasn77:
**Description of the change**

Ensure we always use a string and prevent accidental usage of incorrect type
(such as using true or false as values)

**Benefits**

Incorrect value type would lead to templating errors and would require additional escaped quotes. This bypasses those issues and should make it less error-prone to users. 

**Possible drawbacks**

N/A

**Applicable issues**

- Mentioned in to #137

**Additional information**

Pretty simple change

**Checklist** 

### DIFF
--- a/charts/plex/Chart.yaml
+++ b/charts/plex/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 1.20.2.3402
 description: Plex Media Server
 name: plex
-version: 2.1.0
+version: 2.1.1
 keywords:
   - plex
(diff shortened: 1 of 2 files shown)


onedr0p:
@yasn77 you'll need to bump the patch version in Chart.yaml

yasn77:
> @yasn77 you'll need to bump the patch version in Chart.yaml

Oops... I actually did, just forgot to force push to my branch

onedr0p (approved):

PR #414, opened 2020-12-18
[common] allow code-server to run with additional args

billimek:
**Description of the change**

Enable code-server addon to run with additional configured runtime arguments

**Benefits**

Enables additional runtime features, such as: 

```
      --user-data-dir        Path to the user data directory.
      --extensions-dir       Path to the extensions directory.
      --list-extensions      List installed VS Code extensions.
      --force                Avoid prompts when installing VS Code extensions.
      --install-extension    Install or update a VS Code extension by id or vsix. The identifier of an extension is `${publisher}.${name}`.
                             To install a specific version provide `@${version}`. For example: 'vscode.csharp@1.2.3'.
      --enable-proposed-api  Enable proposed API features for extensions. Can receive one or more extension IDs to enable individually.
```

As configured currently, the code-server addon will never retain state for code-server extensions, settings, etc as these are all persisted in `~/.local`.  This can be overridden with the `--user-data-dir` argument to reference a persisted location (e.g. `/config/.vscode`)

**Checklist** 

### DIFF
--- a/charts/common/templates/addons/code-server/_container.tpl
+++ b/charts/common/templates/addons/code-server/_container.tpl
@@ -30,7 +30,6 @@ args:
 - "--port"
 - "{{ .Values.addons.codeserver.service.port.port }}"
 - {{ .Values.addons.codeserver.workingDir | default (first .Values.addons.codeserver.volumeMounts).mountPath }}
-- {{ .Values.addons.codeserver.workingDir | default (first .Values.addons.codeserver.volumeMounts).mountPath }}
 {{- with .Values.addons.codeserver.volumeMounts }}
 volumeMounts:
   {{- toYaml . | nindent 2 }}
--- a/charts/common/templates/addons/code-server/_container.tpl
+++ b/charts/common/templates/addons/code-server/_container.tpl
@@ -24,13 +24,12 @@ ports:
   containerPort: {{ .Values.addons.codeserver.service.port.port }}
   protocol: TCP
 args:
+{{- range .Values.addons.codeserver.args }}
+- {{ . | quote }}
+{{- end }}
 - "--port"
 - "{{ .Values.addons.codeserver.service.port.port }}"
-- "--auth"
-- "none"
-{{- with .Values.addons.codeserver.args }}
-{{ . }}
-{{- end }}
+- {{ .Values.addons.codeserver.workingDir | default (first .Values.addons.codeserver.volumeMounts).mountPath }}
 - {{ .Values.addons.codeserver.workingDir | default (first .Values.addons.codeserver.volumeMounts).mountPath }}
 {{- with .Values.addons.codeserver.volumeMounts }}
 volumeMounts:
--- a/charts/common/templates/addons/code-server/_container.tpl
+++ b/charts/common/templates/addons/code-server/_container.tpl
@@ -28,6 +28,9 @@ args:
 - "{{ .Values.addons.codeserver.service.port.port }}"
 - "--auth"
 - "none"
+{{- with .Values.addons.codeserver.args }}
+{{ . }}
+{{- end }}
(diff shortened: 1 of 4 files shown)


charts/common/templates/addons/code-server/_container.tpl
```diff
@@ -28,6 +28,9 @@ args:
 - "{{ .Values.addons.codeserver.service.port.port }}"
 - "--auth"
 - "none"
+{{- with .Values.addons.codeserver.args }}
+{{ . }}
+{{- end }}
```
billimek, line 33: @bjw-s do you suppose there is a better/cleaner way to achieve this here?

bjw-s: I was thinking maybe you could replace the `args` section with this:

```yaml
args:
{{- range .Values.addons.codeserver.args }}
- {{ . | quote }}
{{- end }}
- "--port"
- "{{ .Values.addons.codeserver.service.port.port }}"
- {{ .Values.addons.codeserver.workingDir | default (first .Values.addons.codeserver.volumeMounts).mountPath }}
```

That way we only force the port and the startup folder, and leave everything else configurable

charts/common/values.yaml
```diff
@@ -314,6 +314,9 @@ addons:
       pullPolicy: IfNotPresent
       tag: 3.7.4
 
+    # Add additional args
+    args: []
```
bjw-s, line 318:
```suggestion
    # Set codeserver command line arguments
    args:
      - --auth
      - none
```

onedr0p:
@billimek you'll need to bump the version in the common-test chart as well. I ran into this with the 2.0.0 release, hence all the patch bumps lol

bjw-s:
> @billimek you'll need to bump the version in the common-test chart as well. I ran into this with the 2.0.0 release, hence all the patch bumps lol

Really!? 🤔 That seems undesirable... I thought it would only need to be updated when there's an actual change to that chart (e.g. new or modified tests)

onedr0p:
When we publish a new version of common it tries to release common-test too

bjw-s:
I tried going through the CI logs and the chart-releaser code, but can't really find a reason why it would do that. 
Maybe we try it one more time without bumping common-test, if it still borks we'll fix that for now (yay for common-2.1.1 😆), and start looking at a more permanent fix?

onedr0p:
Sounds good to me!

charts/common/templates/addons/code-server/_container.tpl
billimek: much better approach, thank you!

charts/common/templates/addons/code-server/_container.tpl
```diff
@@ -24,13 +24,12 @@ ports:
...
 - "--port"
 - "{{ .Values.addons.codeserver.service.port.port }}"
-- "--auth"
-- "none"
-{{- with .Values.addons.codeserver.args }}
-{{ . }}
-{{- end }}
+- {{ .Values.addons.codeserver.workingDir | default (first .Values.addons.codeserver.volumeMounts).mountPath }}
 - {{ .Values.addons.codeserver.workingDir | default (first .Values.addons.codeserver.volumeMounts).mountPath }}
```
onedr0p: Is this line supposed to be duplicated?

billimek: no; fixed in https://github.com/k8s-at-home/charts/pull/414/commits/52829d78616fb2c5b596e41cb16e6fd81f677e6b

onedr0p: just saw that :)

onedr0p:
Might want to want for https://github.com/k8s-at-home/charts/pull/415 to be merged first

bjw-s (approved):
Changes looking good 🙂

PR #425, opened 2020-12-21
[resilio-sync]: new chart

salekseev:
**Description of the change**

Adding a chart for resilio-sync based on [linuxserver image](https://github.com/orgs/linuxserver/packages/container/package/resilio-sync).

**Checklist** 

### DIFF
--- a/charts/resilio-sync/.helmignore
+++ b/charts/resilio-sync/.helmignore
@@ -1,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
+*.tmp
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
+# OWNERS file for Kubernetes
+OWNERS
--- a/charts/resilio-sync/.helmignore
+++ b/charts/resilio-sync/.helmignore
@@ -19,5 +19,6 @@
 .project
 .idea/
 *.tmproj
+.vscode/
 # OWNERS file for Kubernetes
 OWNERS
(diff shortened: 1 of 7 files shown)


charts/resilio-sync/values.yaml
```diff
@@ -0,0 +1,94 @@
+# Default values for resilio-sync.
+
+image:
+  repository: ghcr.io/linuxserver/resilio-sync
+  pullPolicy: IfNotPresent
+  tag: 2.7.2.1375-ls75
```
onedr0p, line 6: I think lsio has version tags, see if one like this exists `version-2.7.2.175`

charts/resilio-sync/values.yaml
```diff
@@ -0,0 +1,94 @@
...
+    ## If set to "-", storageClassName: "", which disables dynamic provisioning
+    ## If undefined (the default) or set to null, no storageClassName spec is
+    ##   set, choosing the default provisioner.  (gp2 on AWS, standard on
+    ##   GKE, AWS & OpenStack)
+    # storageClass: "-"
+    # accessMode: ReadWriteOnce
+    # size: 1Gi
+    ## Do not delete the pvc upon helm uninstall
+    # skipuninstall: false
+    # existingClaim: ""
```
onedr0p, line 77: You can omit these, config and sync should be enough to get going with persistent, people can always choose to add whatever they want here

charts/resilio-sync/.helmignore
```diff
@@ -0,0 +1,23 @@
...
+.svn/
+# Common backup files
+*.swp
+*.bak
+*.tmp
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
```
onedr0p, line 21: Add `.vscode/` under this line

onedr0p (changes requested):

charts/resilio-sync/values.yaml
salekseev: Seems lsio default config comes with both `/downloads` and `/sync` pre-configured (https://github.com/linuxserver/docker-resilio-sync/blob/master/root/defaults/sync.conf#L8-L9) so I figured I'd provide a fully working example here.

onedr0p: Lsio loves doing weird things, but this is fine 😅

onedr0p (approved):

PR #427, opened 2020-12-21
[searx] new chart

auricom:
This chart is based on the docker-compose repository, on searx gitub organization.

Months ago, I translated it in kubernetes static yaml template files, the configuration being declared in ConfigMaps, in order to have a stateless kubernetes deployment.

This is my attempt to make a helm chart using the k8s-at-home common library.

**Overview**

This chart will create a pod with 5 containers
Configuration is done using Secret and ConfigMap.
Persistent storage is not required.
The user will need to provide values to variables (documented on README.md).

**Notes**
Using the default values provided by values.yaml, the pod will have two container that listen to TCP/80 : searx and caddy reverse proxy ; the problem is that there is no service that will listen to TCP/80 inside searx container. It is the only way I found to have default service to work with this setup.
The workaround would be that Caddy reverse proxy will become the main container of this chart, but I found this inconvenient.

Please squash the commits before merging into master

**Checklist**


auricom:
@carpenike you may want to test this chart as said on discord

carpenike (approved):
LGTM

PR #429, opened 2020-12-23
[home-assistant] allow setting a base64-encoded SSH private key

onedr0p:
**Description of the change**

Allows setting a Base64-encoded SSH private key

**Benefits**

Using sealed-secrets and `.env` I need to pass this var in as a single line, SSH private keys are multiline

**Possible drawbacks**

None

**Applicable issues**

- fixes #

**Additional information**

**Checklist** 

### DIFF
--- a/charts/home-assistant/templates/secret.yaml
+++ b/charts/home-assistant/templates/secret.yaml
@@ -1,4 +1,4 @@
-{{- if .Values.git.deployKey }}
+{{- if or .Values.git.deployKey .Values.git.deployKeyBase64 }}
 apiVersion: v1
 kind: Secret
 metadata:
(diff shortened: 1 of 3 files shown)


charts/home-assistant/values.yaml
```diff
@@ -18,12 +18,19 @@ service:
...
 # securityContext:
 #   privileged: true
 
-git:
-  deployKey: ""
+# Allow access a Git repository by passing in a private SSH key
+# git:
+#   # Raw SSH private key
+#   deployKey: ""
+#   # Base64-encoded SSH private key
```
bjw-s, line 32:
```suggestion
#   # Base64-encoded SSH private key. When both variables are set, the raw SSH key takes precedence.
```

bjw-s (commented):
Small suggestion, otherwise LGTM!

charts/home-assistant/templates/secret.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.git.deployKey }}
+{{- if or .Values.git.deployKey .Values.git.deployKeyBase64 }}
```
bjw-s, line 1: This won't work with the new `values.yaml` :( We need a nested `if` then:

```yaml
{{- if .Values.git }}
{{- if or .Values.git.deployKey .Values.git.deployKeyBase64 }}
...
{{- end }}
{{- end }}
```

PR #464, opened 2021-01-02
[TVheadend] - New chart contribution

mr-onion-2:
New chart for [TVheadend](https://github.com/tvheadend/tvheadend), using the common library.


onedr0p (approved):

onedr0p:
Thanks for another chart!

charts/tvheadend/Chart.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: v2
+appVersion: release-4.2
```
onedr0p, line 2: This likely needs to be 4.2

onedr0p (changes requested):

charts/tvheadend/Chart.yaml
mr-onion-2: Ah ok, I thought it should match the image tag

mr-onion-2: The image version tag is indeed "release-4.2". Happy to change though, please confirm :)

onedr0p: I think it should be the app version, not the image version. LSIO tagging is dumb

mr-onion-2: Ahha! Thanks - learn something new every day

charts/tvheadend/Chart.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: v2
+appVersion: 4.2
```
onedr0p, line 2: Actually looking at LSIO image tags for this image, I would roll with `appVersion: latest` here.

charts/tvheadend/values.yaml
```diff
@@ -0,0 +1,30 @@
+# Default values for tvheadend.
+
+image:
+  repository: linuxserver/tvheadend
+  pullPolicy: IfNotPresent
+  tag: release-4.2
```
onedr0p, line 6: Actually looking at LSIO image tags for this image, I would roll with `tag: latest` here.

onedr0p (changes requested):
LSIO has the strangest way to tag, I would use `latest` on this. It is not the first chart we've had to do this with. Some applications developers and LSIO maintiners just don't follow proper semver. Usually I have been building our own images at https://github.com/k8s-at-home/container-images but this isn't even properly tagged by the developer hence LSIO dumb tags

onedr0p:
Also worth nothing that v4.2 is very old but it looks like the developer is pretty active in commits in the repo.

charts/tvheadend/Chart.yaml
mr-onion-2: No probs! Done

charts/tvheadend/values.yaml
mr-onion-2: Done

onedr0p:
Perfecto!

onedr0p (approved):

mr-onion-2:
> Also worth nothing that v4.2 is very old but it looks like the developer is pretty active in commits in the repo. They just aren't tagging new releases.

Cheers yeh I noticed that. Waiting for [#1348](https://github.com/tvheadend/tvheadend/pull/1348) to be merged to allow Plex integration

onedr0p:
So is this pretty much the same thing as xTeVe?

https://github.com/xteve-project/xTeVe

mr-onion-2:
Similar yeh, as far I understand. Tvheadend supports physical tuners, and with tvhproxy the channels can be added to Plex DVR. The PR I referenced above will remove the need to use tvhproxy.

Flole998:
> Also worth nothing that v4.2 is very old but it looks like the developer is pretty active in commits in the repo. They just aren't tagging new releases.

The problem is that the last active developer left right when the new release was supposed to happen soon, so I took over so the project isn't completely abandoned. I still need to find some time to figure out how a proper release is made (so the build system picks it up, official builds are created and so on) and that's why there is no release. Long story short: Tvheadend needs more active developers ;)

onedr0p:
ThankS for the update @Flole998 ! We will be sure to update the chart once you release a new version :)

PR #475, opened 2021-01-06
[Mosquitto] Update Mosquitto to support configurable ports

brainstain:
**Description of the change**

Updated the service to have configurable ports for users that wish to run on non-standard. 

**Benefits**

Users can configure the ports that they want the service to expose

**Possible drawbacks**

**Applicable issues**

**Additional information**
Tested on microk8s 

**Checklist** 

### DIFF
--- a/charts/mosquitto/Chart.yaml
+++ b/charts/mosquitto/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "1.6.12"
 description: Eclipse Mosquitto - An open source MQTT broker
 name: mosquitto
-version: 0.5.0
+version: 0.6.0
 keywords:
   - message queue
   - MQTT
--- a/charts/mosquitto/templates/NOTES.txt
+++ b/charts/mosquitto/templates/NOTES.txt
@@ -1,6 +1,6 @@
 ** Please be patient while the chart is being deployed **
 
-Mosquitto can be accessed within the cluster on port 1883 at {{ template "mosquitto.fullname" . }}.{{ .Release.Namespace }}.svc.cluster.local
+Mosquitto can be accessed within the cluster on port {{ .Values.service.port }} at {{ template "mosquitto.fullname" . }}.{{ .Release.Namespace }}.svc.cluster.local
 
 To access for outside the cluster, perform the following steps:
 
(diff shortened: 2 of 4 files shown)


onedr0p:
This might be confusing for people if you don't show how to change the non standard port of the application, is it a cli arg, environment variable?

brainstain:
@onedr0p It is configurable via the values.yaml file and the defaults are in place.  This change would not impact anyone utilizing the chart.  https://github.com/k8s-at-home/charts/pull/475/files#diff-59790794f32c9f4d795399384f2551f17017e20fe5fd422b3c2cea5506f5eb1fR37-R38

Did I misunderstand your suggestion?

onedr0p:
The application by default uses port 1883 and 9001. If someone wanted to use different ports the application won't start because the ports in the container would still be 1883 and 9001.

bjw-s:
> The application by default uses port 1883 and 9001. If someone wanted to use different ports the application won't start because the ports in the container that the application is using would still be 1883 and 9001.

Yep, also, the container ports are also still listening on 1883 and 9001: https://github.com/brainstain/charts/blob/5eecbb6f14fcc5625defa4d1dae1d59d2c2dd0b7/charts/mosquitto/templates/statefullset.yaml#L52

brainstain:
Odd, this works on my deployment.  My understanding is since it is a named port, k8s will bind the exposed service port to  the one that is exposed via the container.  I'll do more testing to verify.

brainstain:
Container ports are informational only, per the k8s docs https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/: 

List of ports to expose from the container. Exposing a port here gives the system additional information about the network connections a container uses, but is primarily informational. Not specifying a port here DOES NOT prevent that port from being exposed. Any port which is listening on the default "0.0.0.0" address inside a container will be accessible from the network. Cannot be updated.
--
I can add a target port to the service.

bjw-s:
Even if it is only informative and doesn't affect the operation of the deployment, it would still be nice if the information was correct. Personally I would like to see that the service ports are 'dynamic' in the statefulset too.

brainstain:
Okay, I did more research into this, and the implementation is correct.  If you look at the service, it includes the targetport as a named entity: default & websocket respectively.  https://github.com/brainstain/charts/blob/5eecbb6f14fcc5625defa4d1dae1d59d2c2dd0b7/charts/mosquitto/templates/service.yaml#L21

These are named in the statefulset: https://github.com/brainstain/charts/blob/5eecbb6f14fcc5625defa4d1dae1d59d2c2dd0b7/charts/mosquitto/templates/statefullset.yaml#L51

What this does is map the port defined in the service to the port listened to by the container.  I don't want to update the listed container port because that would be misleading if someone wanted to use a type of NodePort to do port forwarding

brainstain:
FWIW, I tested this on multiple different types of ports and is working as expected.

onedr0p:
Ok that makes sense! Thanks for the follow up, we can get this merged :)

PR #485, opened 2021-01-10
[Mosquitto] upgrade to 2.0.4

lazyoldbear:
Upgraded from 1.6.x to 2.x, with minimal changes to the default config to allow it work out of the box in insecure way. 

Signed-off-by: Ingvarr Zhmakin

### DIFF
--- a/charts/mosquitto/Chart.yaml
+++ b/charts/mosquitto/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "2.0.4"
 description: Eclipse Mosquitto - An open source MQTT broker
 name: mosquitto
-version: 0.6.0
+version: 0.7.0
 keywords:
   - message queue
   - MQTT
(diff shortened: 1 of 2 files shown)


lazyoldbear:
Apologies, broke persistence. Corrected just now.

onedr0p:
This looks like maybe a potential breaking change, should the chart version be bumped to `1.0.0`?

lazyoldbear:
Well, based on what I see in Mosquitto changes, they are more security cleanup. And the default config is "just work", with an advantage of more relevant starting point in configuration. 
Besides, 1.x was in 0.x of the chart. The changes do not really look that dramatic. 
Regardless, I am happy to live with any version policy.

onedr0p:
Looking into it more, you are right, we don't need a 1.0.0 release. Let's merge it as is :)

onedr0p (approved):

PR #492, opened 2021-01-12
[bitwardenrs] Add option to change web port

### LINKED ISSUE #491: [bitwardenrs] Allow for specifying web port for unprivileged pod
# Details

**Helm chart name:** bitwardenrs

**Describe the solution you'd like:**
Allow for the changing of the web ui port, adding the following to the values.yaml:
```yaml
bitwardenrs:
  gui:
    # If you set a different port here, you must also provide it under env
    port: 80
```

**Additional Information:**
This allows users to run the pod unprivileged, as non-root accounts cannot bind ports below 1024.

WojoInc:
Signed-off-by: TJ Wesolowski <wojoinc@pm.me>

**Description of the change**

Adds an option to the values.yaml:
```yaml
bitwardenrs:
  gui:
    # If you set a different port here, you must also provide it under env
    port: 80
```

**Benefits**

When combined with the following:
```yaml
env:
  ROCKET_PORT: "<port>"
```
this allows the user to specify a different port for bitwarden to bind to for the WebUI.
This has other benefits, but the most significant is that it allows the user to specify a web port above 1024 when running the container as a non-root user. For example, with the following SecurityContext:
```yaml
securityContext:
  capabilities:
    drop:
      - ALL
  readOnlyRootFilesystem: true
  runAsNonRoot: true
  runAsUser: 65534
```

**Possible drawbacks**

None that I can see so far.

**Applicable issues**

- fixes #491
- Related: dani-garcia/bitwarden_rs#776

**Additional information**

Bumped the minor version, as this should be a backwards-compatible change. Tested an in-place upgrade on my deployment without issues. The default port for the webui has not changed, so existing users should be able to safely upgrade the chart.

**Checklist** 

### DIFF
--- a/charts/bitwardenrs/Chart.yaml
+++ b/charts/bitwardenrs/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: bitwardenrs
 description: Unofficial Bitwarden compatible server written in Rust
 type: application
-version: 1.0.1
+version: 1.1.0
 appVersion: 1.16.3
 keywords:
   - bitwarden
--- a/charts/bitwardenrs/templates/deployment.yaml
+++ b/charts/bitwardenrs/templates/deployment.yaml
@@ -75,7 +75,7 @@ spec:
           {{- end }}
           ports:
             - name: http
-              containerPort: 80
+              containerPort: {{ .Values.bitwardenrs.gui.port }}
               protocol: TCP
             {{- if .Values.bitwardenrs.websockets.enabled }}
             - name: websocket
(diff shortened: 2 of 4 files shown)


WojoInc:
Looks like wget is having issues with DNS in kind?
https://github.com/k8s-at-home/charts/pull/492/checks?check_run_id=1689246462#step:7:249

bjw-s:
> Looks like wget is having issues with DNS in kind?
> https://github.com/k8s-at-home/charts/pull/492/checks?check_run_id=1689246462#step:7:249

I gave the job a retry and it seems to have worked now :-)

bjw-s (approved):
LGTM! 👍

PR #493, opened 2021-01-12
[zwavejs2mqtt] Safer default values.yaml

lazyoldbear:
- Default to non-privileged
- No default device mounts
- `envFrom` example
- Ingress example

### DIFF
--- a/charts/zwavejs2mqtt/Chart.yaml
+++ b/charts/zwavejs2mqtt/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 1.0.0-alpha.2
 description: Fully configurable Zwave to MQTT Gateway and Control Panel
 name: zwavejs2mqtt
-version: 1.0.1
+version: 1.1.0
 keywords:
   - zwave
   - mqtt
--- a/charts/zwavejs2mqtt/Chart.yaml
+++ b/charts/zwavejs2mqtt/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 1.0.0-alpha.2
 description: Fully configurable Zwave to MQTT Gateway and Control Panel
 name: zwavejs2mqtt
-version: 1.0.0
+version: 1.0.1
 keywords:
   - zwave
   - mqtt
--- a/charts/zwavejs2mqtt/values.yaml
+++ b/charts/zwavejs2mqtt/values.yaml
@@ -12,6 +12,12 @@ strategy:
 # https://zwave-js.github.io/zwavejs2mqtt/#/guide/env-vars
 env: {}
   # OZW_NETWORK_KEY:
+  # TZ: Europe/Helsinki
+
+# Get environment from a secret:
+#envFrom:
+#  - secretRef:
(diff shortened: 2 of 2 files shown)


charts/zwavejs2mqtt/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 1.0.0-alpha.2
 description: Fully configurable Zwave to MQTT Gateway and Control Panel
 name: zwavejs2mqtt
-version: 1.0.0
+version: 1.0.1
```
bjw-s, line 5: Since the default value for the privileged mode changes, I would bump this to 1.1.0.
```suggestion
version: 1.1.0
```

bjw-s (changes requested):
Hi, thanks for your contribution! If you could follow [our PR template](https://github.com/k8s-at-home/charts/blob/master/.github/PULL_REQUEST_TEMPLATE.md) the next time, that would really help us with understanding the benefits / potential drawbacks better.

As for this PR, I have left a few remarks, mainly related to duplication of values/documentation.

lazyoldbear:
Reference to common is noted. The reasons I decided to propose that are:

- I did not notice references to `common` right away, my bad. But they might be made more noticeable, like a comment in all values.yaml. Important also to mention that AFAIR, not everything is well documented in common. After a couple of hiccups I just started checking the templates directly. 
- I think I have seen some charts with more extra resources from common, and I reckoned that it would not harm that secrets are better for secrets than plain text, and that ingress is something that can be used. 

But I don't insist. And sorry for messing the template: I was unsure I will get more time, and I felt important to highlight privileged and default mount stuff, for I spent several hours looking for trouble in another place because of that. 
Will make amendments now.

bjw-s:
> * I did not notice references to `common` right away, my bad. But they might be made more noticeable, like a comment in all values.yaml. Important also to mention that AFAIR, not everything is well documented in common.

Yeah, we've seen that as an issue as well, and have raised https://github.com/k8s-at-home/charts/discussions/462. Not much work on it has been done yet though, since we're all a bit time-constrained :D

I will go over the changes in a bit.

charts/zwavejs2mqtt/values.yaml
```diff
@@ -1,5 +1,7 @@
 # Default values for zwavejs2mqtt.
-
+#
+# Please also consider possible values for the `common` library chart.
+#
```
bjw-s, line 4:
```suggestion
# This chart inherits from our common library chart. You can check the default values/options here:
# https://github.com/k8s-at-home/charts/tree/master/charts/common
# 
```

charts/zwavejs2mqtt/values.yaml
```diff
@@ -1,5 +1,7 @@
 # Default values for zwavejs2mqtt.
-
+# This chart inherits from our common library chart. You can check the default values/options here:
+# https://github.com/k8s-at-home/charts/tree/master/charts/common
+#
```
bjw-s, line 4:
```suggestion
#
```

bjw-s (approved):
LGTM

PR #494, opened 2021-01-13
[zigbee2mqtt] Safer default values, improved configuration

lazyoldbear:
Much safer defaults (similar to #493):
- Disabled privileged by default.
- Disabled mounting devices by default. 
- Updated exampled device name to match a more useful case (specific device, not to encourage generic device names)
- Added a comment to check `common` chart.

Additionally:
- Some more options to default configuration, especially useful for Conbee II and/or Home Assistant.

### DIFF
--- a/charts/zigbee2mqtt/values.yaml
+++ b/charts/zigbee2mqtt/values.yaml
@@ -120,7 +120,7 @@ config:
     homeassistant_status_topic: 'homeassistant/status'
 
     # Optional: Enables report feature (see information -> report for more details) (default: false)
-    report: true
+    # report: true
 
     # Optional: Add an elapsed attribute to MQTT messages, contains milliseconds since the previous msg (default: false)
     elapsed: true
--- a/charts/zigbee2mqtt/values.yaml
+++ b/charts/zigbee2mqtt/values.yaml
@@ -103,6 +103,7 @@ config:
     log_level: info
 
     # Optional: ZigBee channel, changing requires re-pairing of all devices. (Note: use a ZLL channel: 11, 15, 20, or 25 to avoid Problems)
+    # default: 11
     channel: 11
 
     # Optional: Baudrate for serial port (default: 115200 for Z-Stack, 38400 for Deconz)
--- a/charts/zigbee2mqtt/values.yaml
+++ b/charts/zigbee2mqtt/values.yaml
@@ -1,7 +1,6 @@
 # Default values for zigbee2mqtt.
 # This chart inherits from our common library chart. You can check the default values/options here:
 # https://github.com/k8s-at-home/charts/tree/master/charts/common
-#
 image:
   repository: koenkk/zigbee2mqtt
   pullPolicy: IfNotPresent
--- a/charts/zigbee2mqtt/values.yaml
+++ b/charts/zigbee2mqtt/values.yaml
@@ -1,5 +1,7 @@
 # Default values for zigbee2mqtt.
-
+# This chart inherits from our common library chart. You can check the default values/options here:
+# https://github.com/k8s-at-home/charts/tree/master/charts/common
+#
 image:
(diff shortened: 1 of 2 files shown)


charts/zigbee2mqtt/values.yaml
```diff
@@ -1,5 +1,7 @@
 # Default values for zigbee2mqtt.
-
+# This chart inherits from our common library chart. You can check the default values/options here:
+# https://github.com/k8s-at-home/charts/tree/master/charts/common
+#
```
bjw-s, line 4:
```suggestion
```

charts/zigbee2mqtt/values.yaml
```diff
@@ -76,16 +81,48 @@ config:
...
+    # Optional: adapter type, not needed unless you are experiencing problems (options: zstack, deconz)
+    # adapter: deconz
 
   advanced:
     network_key: GENERATE
     log_output:
       - console
+    log_level: info
+
+    # Optional: ZigBee channel, changing requires re-pairing of all devices. (Note: use a ZLL channel: 11, 15, 20, or 25 to avoid Problems)
```
bjw-s, line 105:
```suggestion
    # Optional: ZigBee channel, changing requires re-pairing of all devices. (Note: use a ZLL channel: 11, 15, 20, or 25 to avoid Problems)
    # default: 11
```

charts/zigbee2mqtt/values.yaml
```diff
@@ -76,16 +81,48 @@ config:
...
+    channel: 11
```
bjw-s, line 106:
```suggestion
    # channel: 11
```
I suggest commenting this out, since it's offered as an optional config parameter.

charts/zigbee2mqtt/values.yaml
```diff
@@ -76,16 +81,48 @@ config:
...
+
+    # Optional: Add a last_seen attribute to MQTT messages, contains date/time of last Zigbee message
+    # possible values are: disable (default), ISO_8601, ISO_8601_local, epoch (default: disable)
+    last_seen: 'ISO_8601'
+
+    homeassistant_discovery_topic: 'homeassistant'
+    homeassistant_status_topic: 'homeassistant/status'
+
+    # Optional: Enables report feature (see information -> report for more details) (default: false)
+    report: true
```
bjw-s, line 122:
```suggestion
    # report: true
```
I suggest commenting this out, since it's offered as an optional config parameter.

charts/zigbee2mqtt/values.yaml
```diff
@@ -76,16 +81,48 @@ config:
...
+
+    # Optional: Add an elapsed attribute to MQTT messages, contains milliseconds since the previous msg (default: false)
+    elapsed: true
```
bjw-s, line 125:
```suggestion
    # elapsed: true
```
I suggest commenting this out, since it's offered as an optional config parameter.

lazyoldbear:
Sorry for the hard time cleaning my mess today. 
Trying to save the knowledge quickly, with obvious quality... :-)

bjw-s (approved):
LGTM. Will merge once the CI passes (it might fail in the install step, due to the app not being able to start without Zigbee hardware).

lazyoldbear:
It did fail with named device, so I changed the default to ttyACM0. But I would not be surprised if it fails again. No way to tell z2m to stay calm. 
Mocking a device is feasible but is not worth it... 
Perhaps, a workaround would be to harness handling a specific failure.

bjw-s:
> It did fail with named device, so I changed the default to ttyACM0. But I would not be surprised if it fails again. No way to tell z2m to stay calm.
> Mocking a device is feasible but is not worth it...
> Perhaps, a workaround would be to harness handling a specific failure.

That's not possible with the chart-testing tool. But also not really worth it. We know this happens with this chart, and will handle the toil :)

PR #497, opened 2021-01-14
[oauth2-proxy] change image to new home and bump image version

onedr0p:
**Description of the change**

This updates the image repo to https://quay.io/repository/oauth2-proxy/oauth2-proxy and bumps the version to v6.1.1

**Benefits**

I do not see any issues bumping this to v6.1.1, the chart has been working _fine_ for a long time.

**Possible drawbacks**

Could be an edge case where upgrading to v6.x might break some feature in the chart, but we will cross that road when we get to it :)

**Applicable issues**

- fixes #

**Additional information**

**Checklist** 

### DIFF
--- a/charts/oauth2-proxy/Chart.yaml
+++ b/charts/oauth2-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 4.2.0
+version: 4.3.0
 apiVersion: v1
 appVersion: 5.1.0
 home: https://oauth2-proxy.github.io/oauth2-proxy/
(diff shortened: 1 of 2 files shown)


carpenike (approved):

carpenike:
```
[2021/01/14 21:12:50] [logger.go:490] invalid configuration:
  cookie_secret must be 16, 24, or 32 bytes to create an AES cipher, but is 10 bytes
```

looks like the default needs to be changed for CI purposes:

https://github.com/k8s-at-home/charts/blob/ce90cf7af292d5d901f730a7a969da82936cbce1/charts/oauth2-proxy/values.yaml#L12

onedr0p:
Updated, lets see if that helped.

PR #498, opened 2021-01-16
[powerdns] Webserver container port, and startup script fixes

stretched:
**Description of the change**

* Re-add the webserver container port that was removed in a recent change
* Fix post-start script
    - don't trim whitespace between `OPTIONS` and `HOST` arguments to `nc` command
    - retry loop would always terminate after first attempt; re-formatted and re-structured script for clarity

**Benefits**

* Brings back the webserver port, restoring functionality that had worked in older versions (e.g. v2.2.0)
* Fixes the post-start script, saves users from needing create their zone manually

**Possible drawbacks**

* Existing users not overriding `powerdns.domain` will now get a `mydomain.local` zone (re-)created for them on upgrade, and all future restarts
    - should we make the post-start hook conditional on `powerdns.domain` being set, and change the default to empty string?

**Additional information**

**Checklist**


carpenike (approved):
LGTM. Thanks!

Issue #489, opened 2021-01-11
[common] Missing `resources` in values.yaml


bjw-s:
# Details

**Helm chart name and version:** common 2.2.1

**Container name and tag:** N/A

**What steps did you take and what happened:**

`values.yaml` does not contain the `resources` key

**What did you expect to happen:**

`values.yaml` should contain the `resources` key, commented out as an example.

bjw-s:
Fixed with https://github.com/k8s-at-home/charts/commit/44608266dc3af43bea2fffc14743d02a8451e3d0

PR #499, opened 2021-01-17
[common] Allow to override container command

lazyoldbear:
Once we start messing with arguments, it is often convenient to also override the command (which is the container entrypoint). I just hit a use case where that would have been helpful. 

Normal people would never need this, but it is not harmful either. 

Signed-off-by: Ingvarr Zhmakin

### DIFF
--- a/charts/common/CHANGELOG.md
+++ b/charts/common/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.2.2]
+
+### Added
+
+- Allow overriding the main container command.
+
 ## [2.2.1]
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


onedr0p:
This is great, do you think you want to include this as well? 

https://github.com/k8s-at-home/charts/issues/489

Or we can merge this into a new branch for the next release of common.

@bjw-s  what say you? 😀

lazyoldbear:
@onedr0p This sounds like a much better option. A bit embarrassing to make a release just for this. But the contribution guidelines demand bumping the chart version.

bjw-s:
> @bjw-s what say you? 😀

I've created a new `common-next` branch and changed the base branch for this PR. That way we can merge several changes into that before we pack it up into a new release.

charts/common/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: common
 description: Function library for k8s-at-home charts
 type: library
-version: 2.2.1
+version: 2.2.2
```
bjw-s, line 5: Since this is added functionality and not so much a bugfix, a 2.3.0 version would have been the way to go. (https://semver.org/)
```suggestion
version: 2.3.0
```

charts/common/CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.2.2]
```
bjw-s, line 7:
```suggestion
## [2.3.0]
```

lazyoldbear:
Please note: it is arguable that a functionality is really added. 
I'd say you added it when you allowed `args:`. This just makes it complete.

PR #500, opened 2021-01-18
[blocky] refactor config and bump version

onedr0p:
**Description of the change**

- Update version to v0.12
- Sync config.yaml from https://github.com/0xERR0R/blocky/blob/development/docs/config.yml

**Benefits**

**Possible drawbacks**

**Applicable issues**

- fixes #

**Additional information**

**Checklist** 

### DIFF
--- a/charts/blocky/Chart.yaml
+++ b/charts/blocky/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v2
-appVersion: v0.11
+appVersion: v0.12
 description: DNS proxy as ad-blocker for local network
 name: blocky
 version: 4.2.0
--- a/charts/blocky/Chart.yaml
+++ b/charts/blocky/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: v0.11
 description: DNS proxy as ad-blocker for local network
 name: blocky
-version: 4.1.1
+version: 4.2.0
 keywords:
   - blocky
   - dbs
--- a/charts/blocky/Chart.yaml
+++ b/charts/blocky/Chart.yaml
@@ -2,10 +2,11 @@ apiVersion: v2
 appVersion: v0.12
 description: DNS proxy as ad-blocker for local network
 name: blocky
-version: 4.2.0
+version: 5.0.0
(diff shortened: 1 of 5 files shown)


bjw-s (commented):
I feel these changes to the values.yaml could maybe warrant a breaking change...

If somebody hasn't set up values that used to be commented out and are now provided with a default, some very unexpected behavior might occur. For example `clientLookup.upstream` or enabling a whitelist by default

onedr0p:
Doesn't the config get overridden when provided by the user, or does this gets merged by helm as well?

bjw-s:
I believe they all get merged together with the values provided by the user.

So if before I didn't have a whitelist set, and then deploy this chart, all of a sudden I get a whitelist "for free". Same with `config.clientLookup`. If I didn't have it before and don't override it, then all of a sudden I would now get a `config.clientLookup.upstream` value of `udp:192.168.178.1` which could break things.

onedr0p:
That doesn't seem ideal, I feel like this config shouldn't be merged.

I guess the only way todo this would be to make the configmap a string... e.g.

```yaml
config: |
  upstream:
    externalResolvers:
    - 1.1.1.1
    - 9.9.9.9
    conditional:
    ...
```

bjw-s:
Yep, that could work (and may even be the best way to go). That way the config that is defined in user values.yaml is going to be exactly what is passed into the pod.

onedr0p:
@billimek I know originally the chart was made that way, do you see any reason to keep it the way it is currently? Obviously this would be a breaking change.

billimek:
> @billimek I know originally the chart was made that way, do you see any reason to keep it the way it is currently? Obviously this would be a breaking change.

I've been following this PR and think this proposed way makes more sense.  I can't see a good scenario where we would want some blocky config defaults merged with custom config settings.

onedr0p:
Sounds like we have a plan, I'll update this pr to do that and label the version breaking

onedr0p:
Going to go ahead and merge this, thanks for the support @bjw-s :D

bjw-s:
👍🏻

PR #503, opened 2021-01-19
[traefik-forward-auth] Fix artifact hub links in README.md

stretched:
**Description of the change**

Fix artifacthub links in the helm-docs template, and regenerate traefik-forward-auth/README.md

**Checklist** 

### DIFF
--- a/charts/traefik-forward-auth/Chart.yaml
+++ b/charts/traefik-forward-auth/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik-forward-auth
 description: A minimal forward authentication service that provides OAuth/SSO login and authentication for the traefik reverse proxy/load balancer
 type: application
-version: 1.0.4
+version: 1.0.5
 appVersion: 2.2.0
 keywords:
   - traefik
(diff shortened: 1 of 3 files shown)


charts/traefik-forward-auth/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik-forward-auth
 description: A minimal forward authentication service that provides OAuth/SSO login and authentication for the traefik reverse proxy/load balancer
 type: application
-version: 1.0.4
+version: 1.0.5
```
bjw-s, line 5: This should not be required since it's only a documentation change.

stretched: I was following https://github.com/k8s-at-home/charts/blob/master/CONTRIBUTING.md#immutability

> Chart releases must be immutable. Any change to a chart warrants a chart version bump even if it is only changed to the documentation.

I thought README.md was baked into the published chart artifacts.  WDYT?

onedr0p: We've made some adjustments to the way the charts are released in that a version bump shouldn't need to be required for documentation changes. However, I don't think it's something we are looking to enforce though. @bjw-s what do you think?

bjw-s: Agreed, not something we need to enforce, just wanted to share the info :)

onedr0p (approved):

PR #505, opened 2021-01-21
[plex] Add ability to disable and customize probes

mkilchhofer:
**Description of the change**

During the migration from an ubuntu based installation, I needed the ability to disable the Probes, since the listening configs in the migrated config did not serve 200 OK to the kubelet and therefore I was not able to configure plex without `kubectl port-forward`:
```console
$ kubectl exec -ti plex-8f84f474c-r8xx6 -- bash
root@plex-8f84f474c-r8xx6:/# curl http://<Pod IP>:32400/identity
curl: (52) Empty reply from server
root@plex-8f84f474c-r8xx6:/# curl http://127.0.0.1:32400/identity
<?xml version="1.0" encoding="UTF-8"?>
<MediaContainer size="0" claimed="1" machineIdentifier="foo" version="1.20.2.3402-0fec14d92">
</MediaContainer>
```
While here, I thought it would be useful, if we have the ability to freely configure the probes through `values.yaml`

**Benefits**

Probes are now freely configurable and also one can disable it.

**Possible drawbacks**

`-`

**Applicable issues**
`-`

**Additional information**
`-`

**Checklist** 

### DIFF
--- a/charts/plex/Chart.yaml
+++ b/charts/plex/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 1.20.2.3402
 description: Plex Media Server
 name: plex
-version: 2.2.0
+version: 2.3.0
 keywords:
   - plex
 home: https://plex.tv/
--- a/charts/plex/templates/deployment.yaml
+++ b/charts/plex/templates/deployment.yaml
@@ -191,25 +191,18 @@ spec:
             value: "customCertificateDomain={{.Values.certificate.pkcsMangler.setPlexPreferences.customCertificateDomain}}"
 {{- end }}
 {{- end }}
+          {{- if .Values.probes.readiness.enabled }}
           readinessProbe:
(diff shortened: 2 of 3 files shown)


onedr0p:
Looks good, and I just learned what `omit` does :D

onedr0p (approved):

mkilchhofer:
> Looks good, and I just learned about `omit` :D

Haha cool! :D Thanks for merging it.

mkilchhofer:
Oh, I found out why I needed to turn off the probes.
In the network settings i set `Secure connections` to `Required`. With this PR I can simply customize the probes like this in the values file:
```yaml
probes:
  liveness:
    httpGet:
      scheme: HTTPS
  readiness:
    httpGet:
      scheme: HTTPS
  startup:
    httpGet:
      scheme: HTTPS
```
:sunglasses:

PR #509, opened 2021-01-23
[gonic] new chart

onedr0p:
**Description of the change**

music streaming server / subsonic server API implementation

https://github.com/sentriz/gonic

**Benefits**

**Possible drawbacks**

**Applicable issues**

- fixes #

**Additional information**

**Checklist** 

### DIFF
--- a/charts/gonic/values.yaml
+++ b/charts/gonic/values.yaml
@@ -1,0 +1,40 @@
+# Default values for Gonic.
+
+image:
+  repository: sentriz/gonic
+  pullPolicy: IfNotPresent
+  tag: latest
+
+strategy:
+  type: Recreate
+
+service:
+  port:
+    port: 80
+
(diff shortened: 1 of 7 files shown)


charts/gonic/values.yaml
```diff
@@ -0,0 +1,40 @@
+# Default values for Gonic.
+
+image:
+  repository: sentriz/gonic
+  pullPolicy: IfNotPresent
```
bjw-s, line 5: Not sure if defaulting to `Always` here might be better for charts that use a `latest` tag by default?

onedr0p: The devs will be adding semver in their next release, so I can update the image tag in the future. 

https://github.com/sentriz/gonic/pull/112

bjw-s (approved):

PR #510, opened 2021-01-23
[Frigate] adding data pvc and rtmp port to support upcoming 0.8.0 rel…

brujoand:
…ease

Signed-off-by: brujoand <anders@brujordet.no>

**Description of the change**

This adds a service for the rtmp port which is needed for live streaming and sets up a pvc for /data which is where Frigate will store clips and snapshots. 

**Benefits**

This chart will now work with the Home Assistant integration without modification

**Possible drawbacks**

Well, I might have messed something up

**Checklist** 

This chart will probably be moved to the Frigate authors repo some time in the future, so I did not attempt to use the common library.

### DIFF
--- a/charts/frigate/Chart.yaml
+++ b/charts/frigate/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: "0.6.0"
 description: Realtime object detection on RTSP cameras with the Google Coral
 name: frigate
-version: 4.0.1
+version: 4.0.2
 keywords:
   - tensorflow
   - coral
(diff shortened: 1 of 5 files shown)


onedr0p:
I would bump the major version for these changes.

billimek:
I propose that we incorporate these changes as part of the chart migration over to blakeblackshear's repo. 

See https://github.com/blakeblackshear/frigate/issues/156

What do you all think?

onedr0p:
Agreed but we need @blakeblackshear  to migrate https://github.com/blakeblackshear/billimek-charts/tree/master/charts/frigate to https://github.com/blakeblackshear/blakeshome-charts since we don't have anyway to contribute to there.

brujoand:
Sounds good 👍

billimek:
@brujoand FYI, the PR is https://github.com/blakeblackshear/blakeshome-charts/pull/3

PR #515, opened 2021-01-26
Fix indent for additionalVolumeMounts

brianmay:
Without this I get the output:

```
volumeMounts:
nfs: 
  [...]
```

When I expected to see the values indented under the `volumeMounts` section.

### DIFF
--- a/charts/common/Chart.yaml
+++ b/charts/common/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: common
 description: Function library for k8s-at-home charts
 type: library
-version: 2.2.1
+version: 2.2.2
 keywords:
   - k8s-at-home
   - common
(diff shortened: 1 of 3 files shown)


brianmay:
I think this is going to need further work updating the dependencies everywhere. Might need to split into two PRs.

brianmay:
On second thoughts, this value should be a list. In which case it works out OK.

i.e.

```
        volumeMounts:
        - name: data
```

Is fine.

I rather think the value should still be incremented, so will leave this open. But as this is a matter of style, if you disagree, then just close this PR.

bjw-s:
Thank you for submitting a PR! Can you provide an example of the values.yaml that is giving you problems? This seems to be the first time we're hearing about this, and I think there are multiple charts/people using this feature already. 

Also next time, can you maybe fill out the [Pull Request template](https://github.com/k8s-at-home/charts/blob/master/.github/PULL_REQUEST_TEMPLATE.md), that will help us in seeing what the reasoning behind / possible drawbacks are for a PR.

brianmay:
Did you read my 2nd post?

bjw-s:
Closing this PR as suggested by the author.

Issue #519, opened 2021-01-27
[paperless] consumer pod requires that all PVCs be present


carpenike:
@mr-onion-2 -- would like your thoughts on this!

Looks like the values.yaml calls out the additionalVolume which references the PVs. However, those PVs are not required to run per the above persistence block:

https://github.com/k8s-at-home/charts/blob/master/charts/paperless/values.yaml#L29-L56

Wondering what a good layout would be for storage given that some of these files should likely live on an NFS share that persists rather than PVs that can be volatile.

carpenike:
There's other issues with Paperless -- the startup script chowns / chmods all files in its directories. This works fine if using a PV/PVC, but I'm using NFS shares as that's where my printer scans things to. This won't work (operation not permitted).

Opened an issue here: https://github.com/the-paperless-project/paperless/issues/721

mr-onion-2:
Hi there !

I came across that at the weekend actually when I was troubleshooting trying to get it to work with Longhorn. 

As for a good layout, I'm not sure I have any good input for that, but shall have a ponder

carpenike:
@mr-onion-2 Thanks!

I moved over to Teedy instead. Seems to work nicely.

Helm Chart: https://github.com/k8s-at-home/charts/tree/master/charts/teedy
My Deployment: https://github.com/carpenike/k8s-gitops/blob/master/cluster/default/teedy/teedy.yaml

Need to move the "importer" password to a k8s secret and then use the ValueFrom.Secret option for the additional container.

mr-onion-2:
Very nice! I hadn't seen that, I might have to migrate 😁👍
Thanks

What do you think we should we do with this chart?

mr-onion-2:
So I came across a [fork](https://github.com/jonaswinkler/paperless-ng) of Paperless which has a number of improvements including a shiny new front end and machine learning based tag assignment. It also gets around the consumer pod PVC issue. 

There is a discussion going on over at the project's GitHub regarding whether this should be merged into the core or replace it.

I've updated the current [Paperless chart](https://github.com/mr-onion-2/charts/tree/master/charts/paperless) to support the new fork for now in anticipation that it will be merged..

I've updated the major version so do you think it's fair to just update the chart anyway? I guess if anyone wishes to carry on using the 'old' version they still can.

PS. I tried out Teedy but I couldn't find a multi-arch image of the docs-importer. I didn't spend too much time looking though.

bjw-s:
[...]

PR #525, opened 2021-01-31
[Paperless] - Update chart to use the 'Paperless-ng' fork

mr-onion-2:
As described briefly [here](https://github.com/k8s-at-home/charts/issues/519#issuecomment-770238308).

The original Paperless project's codebase has barely changed for years. [Paperless-ng](https://github.com/the-paperless-project/paperless/issues/711) is a fork with a completely new front end, big changes to the core and some nice new features. It feels like a whole new application.

There is some debate as to whether this should be merged into the main project, replace it, or what.

Clearly we don't want a new 'paperless-ng' chart, so I've bumped the major version number on the current chart and changed it to work with the new fork. If paperless-ng gets merged into the main project I'll update the URLs within the chart as necessary. Users who wish to use paperless-og can carry on using the old chart. 

**Checklist** 


bjw-s (approved):
Looks good to me! Thanks for providing these changes! 🏅

PR #529, opened 2021-02-01
[common] Add support for volumeClaimTemplates in statefulset

sennerholm:
**Description of the change**

Adding support for volumeClaimTemplates in statefulset in the common chart

**Benefits**

With this change is possible to use the full power of the statefulset in the common chart, ie it's possible to have unique storage for each instance of a statefulset, and make it possible to scale a statefulset to more than one member.

**Possible drawbacks**

After removing a helm deployment using this functionality the pvc/pv will be kept.

**Applicable issues**

- fixes #

**Additional information**

More information about Statefulset can be found at:
https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#stable-storage

**Checklist** 
- [ x ] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [ x ] Title of the PR starts with chart name (e.g. `[home-assistant]`)
- [ x ] Chart is using our [common](https://github.com/k8s-at-home/charts/tree/master/charts/common) library as a dependency.

The common and common-test chart version is bumped.

### DIFF
--- a/charts/common-test/ci/statefulset-values.yaml
+++ b/charts/common-test/ci/statefulset-values.yaml
@@ -1,0 +1,18 @@
+image:
+  repository: b4bz/homer
+  tag: latest
+  pullPolicy: IfNotPresent
+
+controllerType: statefulset
+
+volumeClaimTemplates:
+  - name: data
+    mountPath: /data
+    accessMode: "ReadWriteOnce"
+    size: 1Gi
+  - name: backup
+    mountPath: /backup
+    subPath: theSubPath
+    accessMode: "ReadWriteOnce"
+    size: 2Gi
+#    storageClass: cheap-storage-class
--- a/charts/common-test/ci/statefulset-values.yaml
+++ b/charts/common-test/ci/statefulset-values.yaml
@@ -15,4 +15,3 @@ volumeClaimTemplates:
     subPath: theSubPath
     accessMode: "ReadWriteOnce"
     size: 2Gi
-#    storageClass: cheap-storage-class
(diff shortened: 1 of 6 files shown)


charts/common/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: common
 description: Function library for k8s-at-home charts
 type: library
-version: 2.3.0
+version: 2.3.1
```
bjw-s, line 5: This can stay at 2.3.0 since version 2.3.0 is not yet released

sennerholm: I had it liked that  first also but was getting an error in the CI check, so I bumped it.

charts/common/values.yaml
```diff
@@ -209,6 +209,19 @@ additionalVolumes: []
...
+# - name: data
+#   mountPath: /data
+#   accessMode: "ReadWriteOnce"
+#   size: 1Gi
+# - name: backup
+#   mountPath: /backup
+#   subPath: theSubPath
+#   accessMode: "ReadWriteOnce"
+#   size: 2Gi
+#  storageClass: cheap-storage-class
```
bjw-s, line 223:
```suggestion
#   storageClass: cheap-storage-class
```

charts/common-test/ci/statefulset-values.yaml
```diff
@@ -0,0 +1,18 @@
...
+  - name: data
+    mountPath: /data
+    accessMode: "ReadWriteOnce"
+    size: 1Gi
+  - name: backup
+    mountPath: /backup
+    subPath: theSubPath
+    accessMode: "ReadWriteOnce"
+    size: 2Gi
+#    storageClass: cheap-storage-class
```
bjw-s, line 18:
```suggestion
```

bjw-s (approved):

PR #530, opened 2021-02-01
[common] Add support for templatified env-variables

sennerholm:
**Description of the change**

Add support for templatified env-variables in the common charts

**Benefits**

Make it possible to be much more dynamic in what values you assign to your environment variables, so it's possible to get for example the "Release-name" into an variable.

**Possible drawbacks**

The env set's is running throught the tpl functions, so any limitation of that is also a limitation of the functionality.

**Applicable issues**

- fixes #

**Additional information**

**Checklist** 
- [] Chart is using our [common](https://github.com/k8s-at-home/charts/tree/master/charts/common) library as a dependency.

Try to not bump the common chart to see if it's breaks. It's not against master and it's already other PRs wish change the version.

### DIFF
--- a/charts/common-test/ci/env-tpl.yaml
+++ b/charts/common-test/ci/env-tpl.yaml
@@ -1,0 +1,11 @@
+image:
+  repository: b4bz/homer
+  tag: latest
+  pullPolicy: IfNotPresent
+
+env: 
+  VALUE: "sampleHardCodedValue" 
+
+envTpl:
+  TEMPLATE_VALUE: "{{ .Release.Name }}-admin"
+  
--- a/charts/common-test/ci/env-tpl.yaml
+++ b/charts/common-test/ci/env-tpl.yaml
@@ -8,4 +8,3 @@ env:
 
 envTpl:
   TEMPLATE_VALUE: "{{ .Release.Name }}-admin"
-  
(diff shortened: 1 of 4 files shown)


charts/common-test/ci/env-tpl.yaml
```diff
@@ -0,0 +1,11 @@
...
+  repository: b4bz/homer
+  tag: latest
+  pullPolicy: IfNotPresent
+
+env: 
+  VALUE: "sampleHardCodedValue" 
+
+envTpl:
+  TEMPLATE_VALUE: "{{ .Release.Name }}-admin"
+
```
bjw-s, line 11:
```suggestion
  
```

charts/common/values.yaml
```diff
@@ -35,6 +35,11 @@ serviceAccount:
 env: {}
 # TZ: UTC
 
+envTpl: {}
+# Variable with values set from templates, example
+# TEMPLATE_VALUE: "{{ .Release.Name }}-admin"
+# With a release name of: demo, the env value will be: demo-admin
+
```
bjw-s, line 42:
```suggestion
## Variables with values set from templates, example
## With a release name of: demo, the example env value will be: demo-admin
envTpl: {}
# TEMPLATE_VALUE: "{{ .Release.Name }}-admin"
```

bjw-s (approved):

PR #532, opened 2021-02-01
[smarter-device-manager] New chart for humble hardware provisioner

### LINKED ISSUE #490: Consider a chart for smarter-device-manager
I have recently discovered a wonderful thing made by Alexandre Ferreira from ARM, called Smarter Device Manager. 
It allows to manage host hardware in a normal way, so that:
- there is no more need to have containers privileged
- there is no way to attempt attaching a USB device to different pods

Information:
- https://gitlab.com/arm-research/smarter/smarter-device-manager/-/tree/master
- https://community.arm.com/developer/research/b/articles/posts/a-smarter-device-manager-for-kubernetes-on-the-edge

I have successfuly utilised it on my Intel NUC in microk8s, works with Home Assistant and zwavejs2mqtt, zigbee2mqtt to set up. 

I suggest to consider creating a chart here. And happy to share my humble manifest yamls if relevant.

lazyoldbear:
**Description of the change**

A new chart for smarter device manager, as discussed in #490.

Please note: after some consideration, I decided to not use `common`: the amount of adjustments is just larger than setting all up from scratch, and other DaemonSet charts in the repo also don't inherit it, which appears acceptable. 

**Benefits**

- Manage hardware resources allocation (two pods don't eat no same dongle).
- Pods decoupled from host resources.
- No more privileged pods for hardware access.

**Possible drawbacks**

The future of the project (little research in ARM) is unclear.

**Applicable issues**

- fixes #490

**Additional information**

**Checklist** 
- [-] Chart is using our [common](https://github.com/k8s-at-home/charts/tree/master/charts/common) library as a dependency.


charts/smarter-device-manager/README.md
```diff
@@ -0,0 +1,91 @@
...
+    nummaxdevices: 20
+  - devicematch: ^video[0-9]*$
+    nummaxdevices: 20
+  - devicematch: ^ttyACM[0-9]*$
+    nummaxdevices: 1
+```
+The above configuration allows up to twenty pods to make use of `/dev/snd`, `/dev/rtc0` and each of `/dev/video[0-9]*` devices. 
+Only one pod at a time can use any of the discovered host `/dev/ttyACM[0-9]*` devices. This means that if a pod has reserved a dongle at `/dev/ttyACM0`, another pod requesting the same device will stay pending.
+
+Please note that only the root of the host `/dev` directory is considered for discovery. Therefore, `by-id` paths will not work and some `udev` rules may be necessary.
```
bjw-s, line 43: Maybe put this in bold to put some extra attention on it, as I think it is important information that people could otherwise miss.

charts/smarter-device-manager/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+The hardware are requested by pods through `resources`, e.g.:
```
bjw-s, line 45:
```suggestion
The hardware is requested by pods through `resources`, e.g.:
```

charts/smarter-device-manager/README.md
```diff
@@ -0,0 +1,91 @@
...
+```yaml
+  resources:
+    limits:
+      smarter-devices/ttyUSB-Conbee-2: 1
+    requests:
+      smarter-devices/ttyUSB-Conbee-2: 1
+```
+
+Common considerations about node affinity apply.
```
bjw-s, line 54: What do you mean with this exactly? I thought that when you use this solution it is no longer required to also use node affinity?

charts/smarter-device-manager/values.yaml
```diff
@@ -0,0 +1,76 @@
...
+  pullPolicy: IfNotPresent
+
+hostNetwork: true
+dnsPolicy: ClusterFirstWithHostNet
+
+devicePluginPath: "/var/lib/kubelet/device-plugins"
+
+#
+# Override default configuration
+#
```
bjw-s, line 13:
```suggestion
# Override default configuration
```

charts/smarter-device-manager/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Installing the Chart
+
+To install the chart with the release name `my-release`:
+
+```console
+helm install --name my-release k8s-at-home/smarter-device-manager
+```
+
+Please consider overrading the default configuration through `values.yaml` file.
```
bjw-s, line 27:
```suggestion
Please consider overriding the default configuration through `values.yaml` file. 
```

bjw-s (changes requested):
Thank you for creating this PR! I have added some small points in my review.

I feel the common library still has some elements that can be used here, but we can experiment with this in a later PR.

charts/smarter-device-manager/README.md
lazyoldbear: Very true. Rearranged a bit to make it more visible.

lazyoldbear: Does it look better to you now?

charts/smarter-device-manager/README.md
lazyoldbear: Oj.

charts/smarter-device-manager/README.md
lazyoldbear: Oj.

charts/smarter-device-manager/README.md
lazyoldbear: Well, it means that the thing effectively works like affinity. Which is very confusing. Removed.

lazyoldbear:
As for the common, I started with it. And it is good for DaemonSet and ConfigMap, but I set a lot of variables empty to disable default stuff. 
And since there might be new default stuff later, and this particular tool is extremely simple, it felt acceptable to just let it be. 
Besides, I'd rather holistically consider it together with intel-gpu thing and node feature beast, which are also non-common-based daemonsets.

lazyoldbear:
Perhaps splitting common into base and common (inheriting base) might be an option. So that base is empty by default, and values get set mostly in common. Then more low level stuff can inherit base. And others get it as before through common.

bjw-s:
> Perhaps splitting common into base and common (inheriting base) might be an option. So that base is empty by default, and values get set mostly in common. Then more low level stuff can inherit base. And others get it as before through common.

Thanks for the feedback... I don't think that is a route we'll be taking in the short term, since that would add even more layers to the common library setup. 

I have created a copy of your code and tried to migrate it to use `common` and I think it covers 85-90% of this use case. It is in [this branch for now](https://github.com/k8s-at-home/charts/tree/sdm-common), can you maybe point me to what you think is critically missing from it? Maybe we see if we can implement that in `common` 2.3.0 (or a later version if the changes are more impactful).

P.S., this is not to try and force the usage of the common library for everything, I am mostly trying to see if there is something critical that we are missing.

lazyoldbear:
Nothing looks wrong from a first glance (apologies, can't look closer later and compare the rendered manifests with my running stuff). 
But I see that you ended up putting mandatory mounts to `values`. This was basically thing I disliked very much, and alternative hacking it in templates is just big and difficult to read. At this point I removed the dependency and reinvented all the wheels. 

So it is definitely a matter of taste, I am totally fine with either approach. My only concern is that the chart appears, for it is the only way I found so far to fill the particular gap.

lazyoldbear:
@bjw-s Allright, took a closer look. You seem to have done more or less I did initially (I just cloned zigsomething2mqwhatever). 
Issues: 
- Service which is not needed .
- Port in DS pods which are not needed.
- `priorityClassName` not passed. Not crucial, but safer to keep. 
- `hostname` not passed. Too lazy to dig if it is really necessary.
- Mandatory mounts go into `values.yaml`, whilst not really be a matter of configuration. 
- `.Values.config` is a list, an empty `[]` may be necessary in `values.yaml`. 

This can all be fixed either through `values.yaml` or through `templates/common.yaml`, but that becomes fat. Do you still like it and want it? :-)

onedr0p:
@lazyoldbear IMO let's give the common library a chance, we can reiterate if there's a need for the chart to be standalone. Not knocking against a standalone chart or the work you did here. It's all appreciated!

lazyoldbear:
Listen, no feelings here. I started with `common` and simply did not like where it led, looked way too fragile. 
But it is your call. Your repo, your choice. :-)
Besides, there was not that much work. Just very quick and dirty capturing of the thing I had running in handmade manifests (which is not very reusable). 

_If_ we go `common` way, how to pursue that? Should I make amendments here, or keep that in sdm-branch?

lazyoldbear:
#533

lazyoldbear:
IMHO, it is slightly hideous. But it does use common. 
Syntax may be very much improved: I don't know no this ninja template stuff, forgive me.
What say you, humans?

lazyoldbear:
The dependency on #533 is loose: it sort of works now, and will keep doing that, just will also get priority on common 2.3.0.

bjw-s:
common 2.3.0 has been submitted for review, so should be released soon. I will make getting this PR merged a priority once that's live.

lazyoldbear:
No rush, but there seems to be no strict dependency: priority will not be noticeable in the majority of home installations. 

As a consideration for future, it might be beneficial to write a little note about how hardware gets attached to containers. There is intel GPU thing, node-feature-discovery, smarter device manager. And also an old practice of privileged containers with automation tools directly using host resources. And the topic is not that complicated. 

However, there is a gap in my understanding. I still did not find a nice enough way to run X11-like GUI in stateless way.

This all, however, is a sidetrack.

lazyoldbear:
Ehm.
Is that me, or the robot?.. Can't get what fails here.

lazyoldbear:
Oj.
A moment.

lazyoldbear:
At least the state from the chart has now converged to the state I am running at home. 
More attention might be necessary in future, as well as some consolidation of hardware stuff, as mentioned already.

charts/smarter-device-manager/values.yaml
```diff
@@ -0,0 +1,72 @@
...
+# devPath: "/special-dev/"
+# sysPath: "/special-sys/"
+
+envValueFrom:
+  NODE_NAME:
+    fieldRef:
+      apiVersion: v1
+      fieldPath: spec.nodeName
+
+hostname: smarter-device-management
```
bjw-s, line 26: Is this required for the chart to be functional? If it is not, I would suggest removing this.

charts/smarter-device-manager/values.yaml
```diff
@@ -0,0 +1,72 @@
...
+
+priorityClassName: system-node-critical
```
bjw-s, line 28: Is this required for the chart to be functional? If it is not required I would suggest commenting this out by default and putting a comment why a user would want to enable this.

charts/smarter-device-manager/values.yaml
```diff
@@ -0,0 +1,72 @@
...
+
+# DaemonSet labels
+controllerLabels:
+  role: agent
```
bjw-s, line 32: Is this required for the chart to be functional? If it is not required I would suggest commenting this out by default and putting a comment why a user would want to enable this.

charts/smarter-device-manager/values.yaml
```diff
@@ -0,0 +1,72 @@
...
+
+# DaemonSet annotations
+controllerAnnotations: {}
+
+# Pod annotations
+podAnnotations:
+  node.kubernetes.io/bootstrap-checkpoint: "true"
```
bjw-s, line 39: Maybe describe what this annotation does and why it is required.

bjw-s (changes requested):

lazyoldbear:
Those are all valid points, and the starting state was to reproduce what the author suggested. 
I will have to look at it, it will require some reverse engineering to understand.

charts/smarter-device-manager/templates/common.yaml
```diff
@@ -0,0 +1,77 @@
...
+mountPath: /dev
+{{- end -}}
+{{- define "sdm.sys.volumeMount" -}}
+name: host-sys
+mountPath: /sys
+{{- end -}}
+{{- define "sdm.device-plugins.volumeMount" -}}
+name: host-device-plugins
+mountPath: /var/lib/kubelet/device-plugins
+{{- end -}}
```
Ornias1993, line 32: All these should be easily doable with just common chart and values.yaml.
Why are they here?

bjw-s: This is also using the abilities of the `common` library, except setting the required paths in the chart instead of the values.yaml, preventing accidental removal from the additionalVolumes list.

lazyoldbear: Because values.yaml is for configuration. 
And that thing is rigid, not something to alter.

charts/smarter-device-manager/values.yaml
lazyoldbear: This refers to an obsolete feature. Will kill.

charts/smarter-device-manager/values.yaml
lazyoldbear: Did not find any use in the code. Will kill.

charts/smarter-device-manager/values.yaml
lazyoldbear: I'd say this is a good thing, increases the likelihood for this to be started earlier. In some edge cases that will reduce the mess, in others harmless. 
Comment will be added.

charts/smarter-device-manager/values.yaml
lazyoldbear: Does not seem so.

lazyoldbear:
@bjw-s Cleaned up all that you suggested except priority, and also NODE_NAME that got there by accident. 
I hope to avoid that, setting the baseline from the author's suggestions, but it is good to have it tidy. :-)

charts/smarter-device-manager/values.yaml
```diff
@@ -0,0 +1,61 @@
...
+priorityClassName: system-node-critical
+
+# DaemonSet labels
+controllerLabels: {}
+
+# DaemonSet annotations
+controllerAnnotations: {}
+
+podSecurityContext: {}
+  # fsGroup: 2000
```
bjw-s, line 31:
```suggestion
```

These can be removed since they are the same defaults as set in the base chart.

charts/smarter-device-manager/values.yaml
```diff
@@ -0,0 +1,61 @@
...
+  #   memory: 16Mi
+
+
+nodeSelector: {}
+  # If the discovery should be selectively enabled on the cluster nodes, a selector needs to be defined.
+  # smarter-device-manager : enabled
+
+tolerations: []
+
+affinity: {}
```
bjw-s, line 61:
```suggestion
```
These can be removed since they are the same defaults as set in the base chart.

bjw-s (approved):

PR #533, opened 2021-02-02
[common-next] some additional pod properties

lazyoldbear:
**Description of the change**

Just little stuff which happened to be useful in smarter-device-manager (#490).

**Benefits**

Not harming.

**Possible drawbacks**

Not rewarding?..

**Applicable issues**

N/A

**Additional information**

**Checklist**
- [-] (optional) Variables are documented in the README.md

### DIFF
--- a/charts/common/templates/lib/controller/_pod.tpl
+++ b/charts/common/templates/lib/controller/_pod.tpl
@@ -11,8 +11,17 @@ serviceAccountName: {{ include "common.names.serviceAccountName" . }}
 securityContext:
   {{- toYaml . | nindent 2 }}
 {{- end }}
+{{- with .Values.priorityClassName }}
+priorityClassName: {{ . }}
+{{- end }}
+{{- with .Values.schedulerName }}
+schedulerName: {{ . }}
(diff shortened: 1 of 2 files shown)


lazyoldbear:
I would also like to see something like this in pods: 
```yaml
- env:  
  - name: NODE_NAME
    valueFrom: 
      fieldRef:
        apiVersion: v1 
        fieldPath: spec.nodeName
```
What would be the _nice_ way of doing that? Can only think of clumsy structures.

bjw-s:
> What would be the _nice_ way of doing that? Can only think of clumsy structures.

With the current setup I think only more or less clumsy solutions are possible 🙁 I will think about this some more as well.
Do you want have this PR wait on this?

lazyoldbear:
Well, this PR is for trivial additions, and more complicated env is more to think about. 
I'm hesitant to suggest any specific approach there, but I'd like this to be considered. Opened #538.

bjw-s (approved):
I've slightly modified the `values.yaml` comments, will merge once that CI job finishes.

PR #543, opened 2021-02-04
[lazylibrarian] Move to Common

### LINKED ISSUE #527: [LazyLibrarian] Move to Common Library
LazyLibrarian is a quite simple chart it looks like.
Yet it's not yet using the common chart.

Ornias1993:
**Description of the change**

- This PR moves the LazyLibrarian chart to use the common dependency chart
- Updates the chart to use latest docker container
- and alters documentation accordingly

**Benefits**

- Moving to the common chart should decrease maintenance load on the maintainers

**Possible drawbacks**

- It containers a few breaking changes, which are documented accordingly in the readme

**Applicable issues**

- fixes #527

**Additional information**

**Checklist** 


charts/lazylibrarian/README.md
```diff
@@ -1,4 +1,4 @@
-# LazyLibrarian helm chart
+# LazyLibrarian
 
 This is a helm chart for [LazyLibrarian](https://gitlab.com/LazyLibrarian/LazyLibrarian.git) based on the [container image provided by LinuxServer.io](https://hub.docker.com/r/linuxserver/lazylibrarian/).
```
bjw-s, line 3:
```suggestion
This is a helm chart for [LazyLibrarian](https://gitlab.com/LazyLibrarian/LazyLibrarian.git).

**This chart is not maintained by the upstream project and any issues with the chart should be raised [here](https://github.com/k8s-at-home/charts/issues/new/choose)**
```

charts/lazylibrarian/values.yaml
```diff
@@ -1,153 +1,42 @@
...
-  annotations: {}
-  # The name of the service account to use.
-  # If not set and create is true, a name is generated using the fullname template
-  name: ""
-
-podAnnotations: {}
-
-podSecurityContext: {}
-  # fsGroup: 2000
+  tag: version-047f91af
```
bjw-s, line 6: Is this version the same as `appVersion` in `chart.yaml`? May be better to keep them in sync

charts/lazylibrarian/values.yaml
```diff
@@ -1,153 +1,42 @@
...
-  #   memory: 128Mi
-  # requests:
-  #   cpu: 100m
-  #   memory: 128Mi
+env: {}
+  # TZ: UTC
+  # PUID: 1001
+  # PGID: 1001
+  # DOCKER_MODS=linuxserver/mods:lazylibrarian-ffmpeg
+  # DOCKER_MODS=linuxserver/calibre-web:calibre
```
bjw-s, line 20: I don't think you can have the same env var twice, the second one would overwrite the first. Not sure about the lsio syntax to load multiple docker mods though.

bjw-s (changes requested):

bjw-s:
Thank you for taking the time to create this PR! I've left a few small comments, but in general it looks good!

charts/lazylibrarian/values.yaml
Ornias1993: App version has not changed, latest release is still 1.7.2

charts/lazylibrarian/values.yaml
Ornias1993: Its a comment, just to show users where to put it. 
I don't even know if those work at all, as it isn't part of any testing anyway.

Ornias1993: It looks like in the old code they are just appended after eachother.

Ornias1993: Added a warning comment about using docker mods, to shift blame in case of layer 8 errors.

bjw-s (approved):

PR #547, opened 2021-02-05
[unifi] enhance unifi to log to stdout

billimek:
**Description of the change**

Enhance the unifi helm chart to send server output to `stdout` in addition to the server.log, thanks to [this PR](https://github.com/jacobalberty/unifi-docker/pull/405).

**Benefits**

unifi will send server logs to stdout, making it easier to discover log messages from the unifi controller using your favorite kubernetes logging solution

**Possible drawbacks**

This will probably duplicate and possible deprecate the existing promtail logging shim added to the chart previously.  IMO sending the logs to stdout is a superior approach now that it is an option.

**Checklist** 


billimek:
@michaelkoetter do you have any thoughts on this?  You originally added the promtail integration and I would like to see what you think of this approach.

ishioni:
@billimek One thing you should know about this - it will pipe the controller log, but NOT the mongo log. No way to achieve that unless you're using a separate mongo instance i'm afraid :(

billimek:
> @billimek One thing you should know about this - it will pipe the controller log, but NOT the mongo log. No way to achieve that unless you're using a separate mongo instance i'm afraid :(


Thanks!  I noticed the lack of mongo tailing as well but still saw this as a net gain because right now the default stdout logging is pretty useless. 

Can you think of a reason not to have this set and enabled by default?

ishioni:
Just the mongo issue. Except for that, it should be The Better Way[TM]

bjw-s (approved):

bjw-s:
LGTM except for the conflicts ;)

PR #540, opened 2021-02-03
[unifi] Added conditional, if https svc for captive portal defined, default to https svc

rh-dluong:
**Description of the change**
Add conditional to ingress for captive portal to default to https, otherwise, use http

**Benefits**

Users will be able to use https ingress for captive portal

**Possible drawbacks**
Might create weird behavior for people with https service enabled for captive portal, but aren't actually using it.

**Applicable issues**

- fixes #539 

**Additional information**

**Checklist** 
- [X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [X] Title of the PR starts with chart name (e.g. `[home-assistant]`)
- [X] Chart is using our [common](https://github.com/k8s-at-home/charts/tree/master/charts/common) library as a dependency.
- [X] (optional) Variables are documented in the README.md

I didn't bump up the Chart version actually because I didn't change the chart, only the template.  Can someone test this fix? I currently have my unifi controller up on my one node K8S and my ingress uses the node IP, so I can't deploy it again.

PR #555, opened 2021-02-06
[unifi] Fixed unifi chart breaking from range restriction value on captive ingress

### LINKED ISSUE #539: Ingress for captive portal allow https
# Details

**Helm chart name:**
unifi

**Describe the solution you'd like:**
Have a conditional to define the  service port for https if exists, if not, default to http or none.

https://github.com/k8s-at-home/charts/blob/ef5f8b3451331c48139fbfec16842806defd0418/charts/unifi/templates/captive-ingress.yaml#L37

Right now, it's hard coded, so even if you have the https svc for captive portal configured, the captive portal ingress won't use it.

**Additional Information:**

rh-dluong:
My last PR broke because range seems like it restricted what I could use for values for that section.  I've defined it as a variable at the top of the template and referenced it using that.

**Description of the change**

Defined the following {{- $captivePortalHttps := .Values.captivePortalService.https -}} so I can use it in a conditional correctly.

**Benefits**

Fixes the broken template

**Possible drawbacks**
None that weren't already introduced in the last PR #540 

**Applicable issues**

- fixes  #539

**Additional information**
I was able to test using dry run so I can confirm it'll go through this time.

**Checklist** 

### DIFF
--- a/charts/unifi/Chart.yaml
+++ b/charts/unifi/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 5.14.23
 description: Ubiquiti Network's Unifi Controller
 name: unifi
-version: 1.4.0
+version: 1.4.1
 keywords:
   - ubiquiti
   - unifi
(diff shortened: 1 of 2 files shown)


billimek (approved):
LGTM

PR #557, opened 2021-02-06
[unifi] Allow service port to define captive portal port in deployment like gui and controller service

rh-dluong:
**Description of the change**
Changed the values of the container ports for the captive portal to be dynamic in the deployment like gui and controller.

**Benefits**
It'll allow users using the captive portal in https mode to use standard https port instead of the default 8843.  I couldn't get around the captive portal appending :8843 to my ingress url, messing it up.

**Possible drawbacks**
None that aren't already exist with setting lit like gui and controller service.

**Applicable issues**

- fixes #

**Additional information**

**Checklist** 


billimek (approved):
LGTM

PR #568, opened 2021-02-10
[common] Add ingress unittest and tpl functionality

sennerholm:
**Description of the change**

Add functionality in the ingress handling that checks for a Tpl version of host and secret, and if those exists evaluate those.
Adding some unittest to ensure that both the old and new functionality works. 
**Benefits**

Make it possible to put "tpl" versions of hosts and secret part of the ingress so it's possible to do multiple installs of the same helmcharts in one namespace with only change of the releasename.

**Possible drawbacks**

**Applicable issues**

- fixes #

**Additional information**

**Checklist** 
- [ X ] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [  X ] Title of the PR starts with chart name (e.g. `[home-assistant]`)
- [ X ] (optional) Variables are documented in the README.md


Ornias1993:
Interesting change...
What would also be nice is a tpl option for labels... Considering some ingresses make heavy use of those....

charts/common/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: common
 description: Function library for k8s-at-home charts
 type: library
-version: 2.4.0
+version: 2.4.1
```
Ornias1993, line 5: New feature, so this would be:
```suggestion
version: 2.5.0
```

charts/common/values.yaml
```diff
@@ -178,14 +178,21 @@ ingress:
...
           pathType: Prefix
   tls: []
   #  - secretName: chart-example-tls
+  # Or if you need a dynamic secretname
+  #  - secretNameTpl: '{{ include "common.names.fullname" . }}-ingress'
   #    hosts:
   #      - chart-example.local
+  # Or a tpl that is evaluated
+  #    hostTpl:
+  #      - '{{ include "common.names.fullname" . }}.{{ .Release.Namespace }}.{{ .Values.ingress.domainname }}'
```
Ornias1993, line 195: Maybe double comment out the actual comment?
```suggestion
      ## Or a tpl that is evaluated
      # hostTpl: '{{ include "common.names.fullname" . }}.{{ .Release.Namespace }}.{{ .Values.ingress.domainname }}'
      paths:
        - path: /
          # Ignored if not kubeVersion >= 1.14-0
          pathType: Prefix
  tls: []
  #  - secretName: chart-example-tls
  ## Or if you need a dynamic secretname
  #  - secretNameTpl: '{{ include "common.names.fullname" . }}-ingress'
  #    hosts:
  #      - chart-example.local
  ## Or a tpl that is evaluated
  #    hostTpl:
  #      - '{{ include "common.names.fullname" . }}.{{ .Release.Namespace }}.{{ .Values.ingress.domainname }}'
```

Ornias1993 (changes requested):
Some slight nitpicks, none-the-less a very interesting change! :)

sennerholm:
> Interesting change...
> What would also be nice is a tpl option for labels... Considering some ingresses make heavy use of those....

@Ornias1993 Could you give an example of how you want and I can try to add some code for it?

Ornias1993:
> @Ornias1993 Could you give an example of how you want and I can try to add some code for it?

How I want what? I just say I think it's an interesting option to do the same thing with labels?!
Not everything is about me, if it is something I would need/want myself I would've already PR'ed it. 
Its just a generally usefull feature imho

Ornias1993 (approved):

sennerholm:
> > @Ornias1993 Could you give an example of how you want and I can try to add some code for it?
> 
> How I want what? I just say I think it's an interesting option to do the same thing with labels?!
> Not everything is about me, if it is something I would need/want myself I would've already PR'ed it.
> Its just a generally usefull feature imho

@Ornias1993, Would something like this be good to have?  Or do you think it would be good to have in the key to? 

```
  annotations: 
    kubernetes.io/ingress.class: nginx
    kubernetes.io/tls-acme: "true"
  annotations:
    static.key: '{{ include "common.names.fullname" . }}-annotations'
```

Ornias1993:
> @Ornias1993, Would something like this be good to have? Or do you think it would be good to have in the key to?
> 
> ```
>   annotations: 
>     kubernetes.io/ingress.class: nginx
>     kubernetes.io/tls-acme: "true"
>   annotationsTpl:
>     static.key: '{{ include "common.names.fullname" . }}-annotations'
> ```

Yeah key too would be prefered I think...

bjw-s (approved):

PR #576, opened 2021-02-12
[common] v3.0.0

### LINKED ISSUE #573: [common] lower readiness initialDelaySeconds
# Details

**Helm chart name:** common, as well as misc others.

**Describe the solution you'd like:** Don't prevent traffic from being served for 30 seconds by default.

**Anything else you would like to add:** For readiness probes, `initialDelaySeconds=1` and `periodSeconds=5`, or `initialDelaySeconds=5` if keeping `periodSeconds=10`.

**Additional Information:** Until the readiness check succeeds, the pod will not be used to serve traffic. With the current setup, even if pods start and are ready immediately, their services will not be usable for at least 30 seconds, because of the high readiness initial delay. This strikes me as an odd behavior to use as a default for everything. Failed readiness checks do not cause restarts.

The rationale behind my two options is that
* the first checks at `1, 6, 11, 16 ...`, for a max delay of 5 seconds after the start of the app, and a 0-5 second delay in the common case of an app that starts serving after 0.1-4.9 seconds.
* the second checks at `5, 15 ...`, for a max delay of 10 seconds after the start of the app, and a 0-5 second delay in the common case of an app that starts serving after 0.1-4.9 seconds.
* a theoretical bad solution that only changes `initialDelaySeconds=0` but keeps `periodSeconds=10` would check at `0, 10, 20 ...`, for a max delay of 10 seconds after the start of the app, and a 5-10 second delay in the common case of an app that starts serving after 0.1-4.9 seconds.

AFAIK; readiness initialDelaySeconds should only be raised above the default 0 to either
* optimize cycles as described in the bullets above
* not waste packets if you're certain the app will take at least that long to start up
* avoid hitting a badly-designed app with an early request that will cause it to misbehave

Please let me know if I'm misunderstanding something.

Also thank you for this project! It certainly beats needing to maintain 100 homebrew charts/containers myself. Now it's just 50 or so. ;) 

I may open a few PRs to add new charts for services like Wallabag, TT-RSS, and Wekan etc.; which seem to fit this repo nicely.

bjw-s:
**Description of the change**

- [x] Enable startupProbe by default
- [x] Match default values for Probes with Kubernetes defaults

**Benefits**

Services will be reported as Ready a lot sooner, minimising downtime.

**Possible drawbacks**

N/A

**Applicable issues**

- fixes #573 

**Additional information**

**Checklist** 

### DIFF
--- a/charts/common/CHANGELOG.md
+++ b/charts/common/CHANGELOG.md
@@ -8,11 +8,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- `probes.startupProbe` is now enabled by default. This will allow for quicker `Ready` states on Pods. It will precede the liveness and readiness Probes.
+- `probes.startupProbe` is now enabled by default. This will allow for quicker `Ready` states on Pods. It will precede the liveness and readiness Probes. *Note:* This requires that you are running [Kubernetes 1.16](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/)
 - Aligned our default values for liveness and readiness Probes with the Kubernetes defaults:
-  - `initialDelaySeconds` is now 0
-  - `timeoutSeconds` is now 1
-  - `failureThreshold` is now 3
+  - `initialDelaySeconds` is now `0`
+  - `timeoutSeconds` is now `1`
+  - `failureThreshold` is now `3`
 
 ## [2.5.1]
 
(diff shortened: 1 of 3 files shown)


billimek:
What do you think about adding `kubeVersion` to the `Charts.yaml` file, to set a minimum kubernetes version? 
 [These docs](https://helm.sh/docs/topics/charts/#the-chartyaml-file) show some details.

Maybe something like,

`kubeVersion: ">=1.16.0-0"`

bjw-s:
> What do you think about adding `kubeVersion` to the `Charts.yaml` file, to set a minimum kubernetes version?

Nice touch! I'll put that in :)

onedr0p:
After this gets merged I'd like to include some important notes on the top of the app charts `values.yaml` about how more configuration can be found in the common libraries `values.yaml`

I know we're trying to not duplicate a lot of things from there but should we at least include `ingress` and comment it out? That resource will/should not change much.

onedr0p (approved):

PR #582, opened 2021-02-13
bump charts to common v3

onedr0p:
**Description of the change**

- bump charts using the common library to common v3
- bump each chart major version
- include `ingress.enabled=false` on each chart `values.yaml`
- include "IMPORTANT NOTE" about each app chart using the common library in each chart `values.yaml`
- set `kubeVersion: ">=1.16.0"` each app charts `Chart.yaml`
- bump docker image versions on a few charts
- fix indent consistency

**Checklist** 

### DIFF
--- a/charts/flaresolverr/Chart.yaml
+++ b/charts/flaresolverr/Chart.yaml
@@ -2,18 +2,19 @@ apiVersion: v2
 appVersion: v1.2.3
 description: FlareSolverr is a proxy server to bypass Cloudflare protection
 name: flaresolverr
-version: 1.1.0
+version: 2.0.0
+kubeVersion: ">=1.16.0"
 keywords:
-  - flaresolverr
-  - jackett
+- flaresolverr
+- jackett
 home: https://github.com/k8s-at-home/charts/tree/master/charts/flaresolverr
 sources:
-  - https://github.com/FlareSolverr/FlareSolverr
-  - https://hub.docker.com/r/flaresolverr/flaresolverr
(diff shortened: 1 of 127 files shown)


onedr0p:
100% this will not pass in the CI and we should merge if no issues are found on review.

charts/flaresolverr/Chart.yaml
```diff
@@ -2,18 +2,19 @@ apiVersion: v2
...
 description: FlareSolverr is a proxy server to bypass Cloudflare protection
 name: flaresolverr
-version: 1.1.0
+version: 2.0.0
+kubeVersion: ">=1.16.0"
 keywords:
-  - flaresolverr
-  - jackett
+- flaresolverr
+- jackett
```
bjw-s, line 9: is this somehow related to Jackett? maybe add the cloudflare tag instead?

onedr0p: As far as I know it only works with jackett, i'm pretty sure it's maintained by the same people.

charts/samba/values.yaml
```diff
@@ -22,12 +28,14 @@ service:
...
-  # SHARE1: share1;/share/samba/share1
-  # SHARE2: share2;/share/samba/share2
+ingress:
+  enabled: false
+
+probes:
+  liveness:
+    enabled: false
+  readiness:
+    enabled: false
```
bjw-s, line 38: This will probably need startupProbe disabled as well

PR #585, opened 2021-02-14
[actions] remove updating timestamps in index.yaml

onedr0p:
**Description of the change**

- Flux2 relies less on this field now
- Chart releaser has been updated to generate this timestamp on the next release https://github.com/helm/chart-releaser/issues/90

I think it's safe to remove now even though the new version of the chart releaser action isn't released yet. Renovatebot should notify us of an upgrade to the action.


billimek:
Should we incorporate the reference to the upgraded chart release action into this change?

onedr0p:
Renovatebot _should_ open a PR when the new version is released.

onedr0p:
@billimek added a reference in the workflow.

bjw-s (approved):

PR #595, opened 2021-02-16
Add Helm Docs

onedr0p:
**Description of the change**

Exploring the use of [Helm Docs](https://github.com/norwoodj/helm-docs) with only the common charts

Each chart directory has `README_CHANGELOG.md.gotmpl`, `README_CONFIG.md.gotmpl` and `README.md.gotmpl`

In the script `./hack/gen-helm-docs.sh`

The two files `README_CHANGELOG.md.gotmpl` and `README_CONFIG.md.gotmpl` are specific to application and should not be overwritten. 

The `README.md.gotmpl` can be overwritten and it consumes `README_CHANGELOG.md.gotmpl` and `README_CONFIG.md.gotmpl` as templates.

Generated the `README.md`s with the following:

```console
./hack/gen-helm-docs.sh
```

Generated the `.helmdocsignore` with the following:

```console
./hack/gen-helmdocsignore.sh
```


charts/common/CUSTOM_CONFIG.md.gotmpl
```diff
@@ -0,0 +1,82 @@
+{{- define "custom.custom.configuration.header" -}}
+## Creating a new chart
+{{- end -}}
+
+{{- define "custom.custom.configuration" -}}
+{{ template "custom.custom.configuration.header" . }}
+
+First be sure to checkout the many charts that already use this like [qBittorrent](../qbittorrent/), [node-red](../node-red/) or the many others in this repository.
```
bjw-s, line 8: Maybe we could also refer to the taskfile / script to create a new chart based on the template chart.

charts/common/README.md.gotmpl
```diff
@@ -1,176 +1,147 @@
...
 ```console
-helm install {{ template "chart.name" . }} {{ template "helm.path" . }} -f values.yaml
+helm install {{ template "chart.name" . }} {{ template "custom.helm.path" . }} -f values.yaml
 ```
 {{- end -}}
 
+{{- define "custom.valuesSection" -}}
+## Values
+
+**Important**: When deploying an application Helm chart you add more values from our common library chart [here](https://github.com/k8s-at-home/charts/tree/master/charts/common/)
```
bjw-s:
```suggestion
**Important**: When deploying an application Helm chart you can add more values from our common library chart [here](https://github.com/k8s-at-home/charts/tree/master/charts/common/)
```

hack/templates/README.md.gotmpl
```diff
@@ -0,0 +1,143 @@
...
+
+```console
+helm install {{ template "chart.name" . }} {{ template "custom.helm.path" . }} -f values.yaml
+```
+{{- end -}}
+
+{{- define "custom.valuesSection" -}}
+## Values
+
+**Important**: When deploying an application Helm chart you add more values from our common library chart [here](https://github.com/k8s-at-home/charts/tree/master/charts/common/)
```
bjw-s:
```suggestion
**Important**: When deploying an application Helm chart you can add more values from our common library chart [here](https://github.com/k8s-at-home/charts/tree/master/charts/common/)
```

bjw-s:
Maybe we also need to add some instruction about what we would expect documentation-wise from contributors:
- which files should they populate
- which files are autopopulated

onedr0p:
Added those discussion points to https://github.com/k8s-at-home/charts/discussions/462

PR #596, opened 2021-02-18
[common] 3.1.0 Addition of secret kind

nicholaswilde:
**Description of the change**

- Added the ability to add secrets as a separate resource in addition to the secret being in the `Deployment.yaml` file.

**Benefits**

- Secrets can now be added as a separate resource.

**Possible drawbacks**

- May cause issues with `envFrom` if `secret` and `envFrom.secretRef` are used at the same time.

**Applicable issues**

- None

**Additional information**

- None

**Checklist** 
- [X] (optional) Chart is using our [common](https://github.com/k8s-at-home/charts/tree/master/charts/common) library as a dependency.


bjw-s:
Hi, thanks for taking the time to submit this PR! I am curious what the use case for this would be, since you would be exposing secrets directly/(potentially) publicly in the `values.yaml` file?

Ornias1993:
@bjw-s You make the assumption values.yaml stays on the system or is used in a gitops style system
I think it's a nice addon for very specific niche usecases, but not very usefull for general or gitops use.

charts/common/Chart.yaml
```diff
@@ -11,3 +11,5 @@ home: https://github.com/k8s-at-home/charts/tree/master/charts/common
 maintainers:
 - name: bjw-s
   email: me@bjw-s.dev
+- name: nicholaswilde
+  email: ncwilde43@gmail.com
```
Ornias1993, line 15: I don't think it's best-practice for every contributor to add their contact details here.

Ornias1993 (changes requested):
Nice addon, besides the maintainer change mentioned below.

Also: It might be nice naming PR's relating to what they do, instead of just the version change.

onedr0p:
> **Possible drawbacks**
> 
> * May cause issues with `envFrom` if `secret` and `envFrom.secretRef` are used at the same time.

Maybe adding a unit test for it under https://github.com/k8s-at-home/charts/blob/master/test/charts/common-test_spec.rb could see if there would be an issue.

charts/common/Chart.yaml
onedr0p: @Ornias1993 we're hoping to get @nicholaswilde on-boarded here, so this is fine.

onedr0p:
@nicholaswilde I updated our wiki with some basic info on how to run the unit tests

https://github.com/k8s-at-home/charts/wiki/Common-library:-Unit-tests

charts/common/Chart.yaml
nicholaswilde: > Also: It might be nice naming PR's relating to what they do, instead of just the version change.

Done

nicholaswilde:
> @nicholaswilde I updated our wiki with some basic info on how to run the unit tests as for writing it takes some rudimentary ruby knowledge
> 
> https://github.com/k8s-at-home/charts/wiki/Common-library:-Unit-tests

I'm not versed at all in ruby but I'll give it a try. I'm currently having issues getting `gem-jq` installed per the instructions above.

bjw-s:
I'll put up my devcontainer setup tomorrow morning, hopefully that will make things a bit easier. Wanted to do that tonight, but my 3D printer won my spare time tonight 😅

onedr0p:
> > @nicholaswilde I updated our wiki with some basic info on how to run the unit tests as for writing it takes some rudimentary ruby knowledge
> 
> > 
> 
> > https://github.com/k8s-at-home/charts/wiki/Common-library:-Unit-tests
> 
> 
> 
> I'm not versed at all in ruby but I'll give it a try. I'm currently having issues getting `gem-jq` installed per the instructions above. 

I'm not either and had issue too but setting that environment variable `export RUBYJQ_USE_SYSTEM_LIBRARIES=1` fixed it for me

nicholaswilde:
Okay. so I did a bit of testing with `helm` (not the unit tests because I couldn't get the tests to run correctly), and the only potential problem is if the `envFrom.secretRef[0].name` is the same as the chart name. If the names are different, I don't think it's a problem because a container `envFrom` can take multiple secrets.

```yaml
$ helm template .
...
        envFrom:
        - secretRef:
            name: RELEASE-NAME-common-test
        - secretRef:
            name: RELEASE-NAME-common-test
...
```
```yaml
# values.yaml
...
secret:
  PASSWORD: my-password

envFrom:
 - secretRef:
     name: RELEASE-NAME-common-test
...
```

bjw-s:
As for running the unit tests, I just added the files I use with VS Code (https://github.com/k8s-at-home/charts/commit/bf0126122caec3b65557dfd38dbf5339e05f4844). I'll update the wiki somewhere between work stuff :)

nicholaswilde:
Okay, I think I've managed to implement a unit test for `secret`. The only thing it doesn't do is check `secret` against `envFrom.secretRef` (ae722d5).

charts/common/templates/_secret.tpl
```diff
@@ -0,0 +1,17 @@
...
+  name: {{ include "common.names.fullname" . }}
+  labels:
+    {{- include "common.labels" . | nindent 4 }}
+type: Opaque
+{{- with .Values.secret }}
+stringData:
+  {{- toYaml . | nindent 2 }}
+{{- end }}
+{{- end }}
+
```
bjw-s, line 17:
```suggestion

```
One newline too many

charts/common/values.yaml
```diff
@@ -32,6 +32,9 @@ serviceAccount:
   # If not set and create is true, a name is generated using the fullname template
   name: ""
 
+secret: {}
```
bjw-s, line 35: I would add a small header comment here explaining what this section does (and maybe a small warning about putting clear text passwords in your helm values).

charts/common/README_CHANGELOG.md.gotmpl
```diff
@@ -10,6 +10,12 @@ All notable changes to this application Helm chart will be documented in this fi
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### [3.1.0]
+
+#### Added
+
+- Allow setting secret values through Helm templates.
```
bjw-s, line 17: This could be misconstrued as that the value for `secrets` implements the Helm `tpl` function which isn't the case. Maybe something like this?

```suggestion
- Allow setting secret values through Helm values file.
```

bjw-s:
Just added a few comments, apart from those small points it looks good to me! 👍

nicholaswilde:
> Just added a few comments, apart from those small points it looks good to me! 👍

Thanks for all of the suggestions!

bjw-s (approved):

PR #603, opened 2021-02-21
[radarr][sonarr][lidarr][readarr] Add custom liveness probe for *arr

St0rmingBr4in:
This change adds a custom liveness probe for radarr, sonarr, lidarr and readarr.
This makes Kubernetes a bit more aware of the status of those services. For instance if the database is inaccessible the / endpoint still gives a 200 response but displays `Unable to open database file`. By querying `system/status` endpoint of the api we get the real status of those services.

- [X] Chart version bumped in `Chart.yaml`
- [X] Title of the PR starts with chart name


St0rmingBr4in:
I could also create a common-radarr-sonarr library chart to reduce duplication. What do you think is the change ok as is ?

onedr0p:
> I could also create a common-radarr-sonarr library chart to reduce duplication. What do you think is the change ok as is ?

No need for yet another common library chart 😄 

The only issue I can see with this probe change is `bash` and the location of the `config.xml`. But I think it's fine because it works with LSIO and should work with the k8s@home container images.

Do you want to add something similar to lidarr and readarr while you are at it?

St0rmingBr4in:
Sure I'll add this for lidarr and readarr

charts/sonarr/values.yaml
```diff
@@ -25,6 +25,27 @@ service:
...
+    custom: true
+    ## The spec field contains the values for the default livenessProbe.
+    ## If you selected custom: true, this field holds the definition of the livenessProbe.
+    spec:
+      exec:
+        command:
+        - /usr/bin/env
+        - bash
+        - -c
+        - curl --fail localhost:8989/api/v3/system/status?apiKey=`IFS=\> && while
```
Ornias1993, line 41: Shouldn't this use target service and not the pod?

St0rmingBr4in: What would be the benefit of targeting the service? Since the healthcheck is on the pod itself I don't see why it would be better.

Ornias1993: Shoot, i'm mistaken indeed.

St0rmingBr4in:
I could be using `sed -n "s/.*<ApiKey>\(.*\)<\/ApiKey>/\1/p" /config/config.xml` that would mean I could use `sh` instead of `bash` since `read -d` would not be used. But it would mean `sed` must be installed. Not sure whether it's better to depend on bash or depend on sed, awk or grep.

onedr0p:
Bash is good IMO. As far as I know it's the most compatible

onedr0p (approved):

St0rmingBr4in:
Thank you :)

PR #605, opened 2021-02-21
[valheim] new chart

rust84:
**Description of the change**

Add a chart for deploying a `valheim-server-docker` game server [https://github.com/lloesche/valheim-server-docker](url).

**Benefits**

Used to successfully deploy a Valheim server at home for myself and some friends. I found an existing chart elsewhere that was missing some features I needed such as pod annotations and loadbalancer settings. So I decided to make a chart using the common library which both provides these things and will be easier to maintain.

**Possible drawbacks**

Spending time being a Viking (if that is a drawback?)

**Applicable issues**

n/a

**Additional information**

n/a

**Checklist** 
- [x ] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [ x] Title of the PR starts with chart name (e.g. `[home-assistant]`)
- [ x] (optional) Chart is using our [common](https://github.com/k8s-at-home/charts/tree/master/charts/common) library as a dependency.
- [ x] (optional) Variables are documented in the README.md


charts/valheim/values.yaml
```diff
@@ -0,0 +1,64 @@
...
+    enabled: false
+    mountPath: /config
+
+probes:
+  liveness:
+    enabled: false
+  readiness:
+    enabled: false
+  startup:
+    enabled: false
```
onedr0p: Need a new line at the end

charts/valheim/values.yaml
```diff
@@ -0,0 +1,64 @@
...
+
+image:
+  repository: lloesche/valheim-server
+  pullPolicy: IfNotPresent
+  tag: latest
+
+strategy:
+  type: Recreate
+
+env: {}
```
onedr0p, line 16: I'd include a link to the available environment variables, for example see here

https://github.com/k8s-at-home/charts/blob/master/charts/unifi-poller/values.yaml#L16

onedr0p (changes requested):

charts/valheim/values.yaml
rust84: Updated.

charts/valheim/values.yaml
rust84: Added with latest commit.

onedr0p:
I don't think this will pass the chart install process.

onedr0p (approved):

charts/valheim/Chart.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v2
+appVersion: latest
+description: Valheim dedicated gameserver with automatic update and world backup support
+name: valheim
+version: 2.0.1
```
onedr0p, line 5: This should be 1.0.0 right?

onedr0p (changes requested):

charts/valheim/Chart.yaml
rust84: Ah yes. For context I was testing it out in my own repo and the 2.0.0 was because of switching to using the common template. Starting at 1.0.0 for the k8s-at-home chart makes sense.

onedr0p (approved):

runtman:
<3

Issue #607, opened 2021-02-21
[common] handle volume configmaps/secrets


ishioni:
# Details

**Helm chart name:**
common

**Describe the solution you'd like:**
I need something like
```yaml
volumeFromConfigMap:
  nameSuffix: config
  data: |
    whatever
```
Or a way to be able to compute release names in additionalVolumes. In transmission this is only a minor nuisance, but stuff like mosquitto 2.x or icantbelieveitsnotvaletudo can't move to common, because they can't work without a config file. 

**Anything else you would like to add:**
In my charts i try to make the config present itself as yaml and then try to process it into the target format ( with toJson for example ). It would be great if I could keeo doing that, but have absolutely no idea how to make that happen :(

**Additional Information:**

bjw-s:
Hi, thanks for raising this issue. 

Currently what you wish to accomplish is already possible through the features of the common library. For example, the [Zigbee2MQTT](https://github.com/k8s-at-home/charts/tree/master/charts/zigbee2mqtt/templates) chart already does something like this.

Could you let us know if this approach would work for you?

ishioni:
Yes! This is exactly what I was looking for :) You can close this

bjw-s:
Cool! Great that that would work :) I'll leave the issue open for now as a reminder for me to add some documentation about this to the wiki.

bjw-s:
I've added this example to the wiki: https://github.com/k8s-at-home/charts/wiki/Common-library:-Creating-a-new-chart

PR #609, opened 2021-02-21
[icantbelieveitsnotvaletudo] update chart

ishioni:
**Description of the change**
App update, deprecated all of http support, making it mqtt only, hence the deletions

**Benefits**
The old version was compatible only up to valetudo 0.6.1, and was extremely slow on 0.6 to boot. This update speeds up map generation about ~35x (from almost 10 seconds to a couple hundred miliseconds)

**Possible drawbacks**
Incompatible with older versions of valetudo. People on 0.6 should not upgrade

**Applicable issues**

- fixes #

**Additional information**

**Checklist** 


nicholaswilde:
Can this chart be modified to use the `common` library?

ishioni:
Very much so, but it needs a configmap to run - #607

PR #614, opened 2021-02-22
[icantbelieveitsnotvaletudo] update and port to common

ishioni:
**Description of the change**
App update, port to common, deprecated all of http support, making it mqtt only, hence the deletions

**Benefits**
The old version was compatible only up to valetudo 0.6.1, and was extremely slow on 0.6 to boot. This update speeds up map generation about ~35x (from almost 10 seconds to a couple hundred miliseconds)

**Possible drawbacks**
Incompatible with older versions of valetudo. People on valetudo 0.6 should not upgrade

**Applicable issues**

- fixes #

**Additional information**

**Checklist** 
- [X] (optional) Chart is using our [common](https://github.com/k8s-at-home/charts/tree/master/charts/common) library as a dependency.


bjw-s (approved):
LGTM

PR #620, opened 2021-02-23
[meta] revert non-common chart readmes [ci-skip]

onedr0p:
**Description of the change**

Helm docs ran over these charts by accident, reverting the readmes


bjw-s (approved):

bjw-s:
Approved! Will let you decide on merge timing :-)

PR #628, opened 2021-02-27
[jackett] update jackett to version-v0.17.591

jmmaloney4:
Signed-off-by: Jack Maloney <jmmaloney4@gmail.com>

**Description of the change**

Update image for jackett chart to `version-v0.17.591`.

**Benefits**

**Possible drawbacks**

**Applicable issues**

- fixes #

**Additional information**

**Checklist** 


charts/jackett/values.yaml
```diff
@@ -17,6 +17,8 @@ env: {}
   # TZ: UTC
   # PUID: 1001
   # PGID: 1001
+  ## Allow Jackett to update inside of the container, but might cause outages if turned on.
+  # AUTOUPDATE: false
```
bjw-s, line 21:
```suggestion
```
This is not really the way containers should run software updates. I don't think we should explicitly mention or support this.

bjw-s (changes requested):
Looks good for the most part 👍 Just a small change requested 😄

charts/jackett/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v2
-appVersion: v0.16.2106
+appVersion: 0.17.591
 description: API Support for your favorite torrent trackers
 name: jackett
-version: 7.0.1
+version: 7.1.0
```
bjw-s, line 5:
```suggestion
version: 7.0.2
```
I would be fine with only bumping the patch version for this

charts/jackett/values.yaml
onedr0p: This environment variable isn't doing anything anyways, I don't see it in the source of Jackett.

onedr0p: It's part of LSIO image, which defaults to `AUTOUPDATE=true`... very odd

charts/jackett/Chart.yaml
onedr0p: Agreed, this is not a feature bump

charts/jackett/values.yaml
jmmaloney4: Yes, the default true was the only reason I added it in.

charts/jackett/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v2
-appVersion: v0.16.2106
+appVersion: 0.17.591
```
onedr0p, line 2:
```suggestion
appVersion: v0.17.591
```

PR #630, opened 2021-02-28
[common] Add option to specify image registry to common chart _container.tpl

jmmaloney4:
…ner.tpl

Signed-off-by: Jack Maloney <jmmaloney4@gmail.com>

**Description of the change**

Allow `image.registry` to be specified, which will change the registry the image is pulled from. This is useful for images like the linuxserver.io images which are also pushed to `ghcr.io`, to avoid docker hub rate limiting. 

**Benefits**

**Possible drawbacks**

**Applicable issues**

- fixes #

**Additional information**

**Checklist** 

### DIFF
--- a/charts/common/templates/lib/controller/_container.tpl
+++ b/charts/common/templates/lib/controller/_container.tpl
@@ -3,7 +3,7 @@ The main container included in the controller.
 */ -}}
 {{- define "common.controller.mainContainer" -}}
 - name: {{ include "common.names.fullname" . }}
-  image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+  image: "{{- if .Values.image.registry }}{{ .Values.image.registry }}/{{- end }}{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
   imagePullPolicy: {{ .Values.image.pullPolicy }}
   {{- with .Values.command }}
   command: {{ . }}


bjw-s:
Hi! 👋 Thanks for taking the time to create this PR!

When specifying an image for a chart, you could refer to that image as follows if you want it to reference a different registry:
```yaml
  values:
    image:
      repository: ghcr.io/k8s-at-home/jackett
      tag: v0.17.584
```

Is there a specific reason why you would want to introduce a separate field instead of using the default notation convention?
Adding the extra field also has the downside of breaking image automation tools such as Renovate, since they also follow the default convention.

onedr0p:
I am also curious why you want these as separate fields @jmmaloney4 ?

jmmaloney4:
All/Most of the bitnami charts ([example](https://github.com/bitnami/charts/blob/4c8fb5fe5634e4a69dd4d8cdb663c114bf993961/bitnami/airflow/values.yaml#L6)) have this as a separate field, and it can be convenient for changing the registry, however if it breaks things its not a big deal. I've just been in the habit of switching my images away from docker hub since they implemented rate limiting, and this is something that has made it more convenient. 

Also, `linuxserver.io` images use `ghcr.io` as the default registry in all of their documentation. not sure if it would make sense to switch the charts that use those images to pull from there by default?

onedr0p:
I can agree to switch all the default LSIO images to `ghcr`!

billimek:
I'm not seeing a compelling reason to separate the image repository from the image name as it currently 'works' with them combined.

Is there a functional benefit to having these separated?

FWIW I've only seen this approach taken with the bitnami charts.

nicholaswilde:
I'm not seeing any benefit, either.

jmmaloney4:
Ok sounds good. Thought it might be useful, but if its not, no worries.

PR #643, opened 2021-03-06
[duplicati] Add possibility to add additional volume mounts

rolfberkenbosch:
**Description of the change**

Add the possibility to add additional volume mounts for example nfs.

**Benefits**

With this option you are able to mount also nfs shares to duplicati so that you can backup them.

**Checklist** 

### DIFF
--- a/charts/duplicati/Chart.yaml
+++ b/charts/duplicati/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: v2.0.5.1
 description: Store securely encrypted backups on cloud storage services!
 name: duplicati
-version: 2.0.2
+version: 2.1.0
 keywords:
   - duplicati
 home: https://github.com/k8s-at-home/charts/tree/master/charts/duplicati
--- a/charts/duplicati/Chart.yaml
+++ b/charts/duplicati/Chart.yaml
@@ -11,5 +11,5 @@ sources:
   - https://hub.docker.com/r/linuxserver/duplicati/
   - https://github.com/duplicati/duplicati
 maintainers:
-  - name: skaro13
+  - name: simoncaron
     email: simon.caron@protonmail.com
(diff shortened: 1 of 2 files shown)


Ornias1993:
I personally think this could just as well be ported to the common chart, instead of porting things already available in the common-chart into this one.

rolfberkenbosch:
> I personally think this could just as well be ported to the common chart, instead of porting things already available in the common-chart into this one.

I think you are right, but I don't have te time te refactor the complete helm chart now to use the common one. It's on my todo list!

charts/duplicati/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: v2.0.5.1
 description: Store securely encrypted backups on cloud storage services!
 name: duplicati
-version: 2.0.1
+version: 2.0.2
```
bjw-s, line 5:
```suggestion
version: 2.1.0
```

Even though this is a small feature addition, it's not a bug fix and therefore warrants a minor version bump.

bjw-s (changes requested):
> > I personally think this could just as well be ported to the common chart, instead of porting things already available in the common-chart into this one.
> 
> I think you are right, but I don't have te time te refactor the complete helm chart now to use the common one. It's on my todo list!

I think this is fine for now. There are a number of charts that aren't (yet) based on the common lib that still receive small improvements like these.

Requested a version bump, otherwise LGTM

Ornias1993:
> > I personally think this could just as well be ported to the common chart, instead of porting things already available in the common-chart into this one.
> 
> I think you are right, but I don't have te time te refactor the complete helm chart now to use the common one. It's on my todo list!

@rolfberkenbosch What refactor, if I read the chart right, this should be mostly it: #644

bjw-s (approved):

PR #645, opened 2021-03-07
[helm-docs] Add helm-docs CI

angelnu:
**Description of the change**

Adds a workflow to create a PR with helm-docs updates

**Benefits**

Ensure documentation stays updated

**Possible drawbacks**

More work to integrate changes when submitter did not update documentation on his own fork prior to PR. This happens when submitter forgets to enable actions on his fork.

**Applicable issues**

- NA

**Additional information**

**Checklist** 

### DIFF
--- a/.github/workflows/charts-helm-docs.yaml
+++ b/.github/workflows/charts-helm-docs.yaml
@@ -2,10 +2,6 @@ name: "Charts: helm-docs"
 
 on:
   workflow_dispatch:
-    inputs:
-      chart:
-        description: 'Chart name'     
-        required: false
   push:
     paths:
     - 'charts/**/README.md.gotmpl'
--- a/.github/workflows/charts-helm-docs.yaml
+++ b/.github/workflows/charts-helm-docs.yaml
@@ -38,7 +38,7 @@ jobs:
         token: ${{ secrets.GITHUB_TOKEN }}
         branch: "helm-docs/ci"
         delete-branch: true
-        title: "chore(update helm-docs) [ci-skip]"
+        title: "chore(docs) update helm-docs [ci-skip]"
         signoff: true
         committer: "${{ github.actor }} <${{ github.actor }}@users.noreply.github.com>"
         author: "${{ github.actor }} <${{ github.actor }}@users.noreply.github.com>"
--- a/.github/workflows/charts-helm-docs.yaml
+++ b/.github/workflows/charts-helm-docs.yaml
@@ -1,4 +1,4 @@
-name: "Charts: Lint and test"
+name: "Charts: helm-docs"
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


angelnu:
Example of PR in my fork - https://github.com/angelnu/charts/pull/5

onedr0p:
@angelnu should we use the script in `hack/gen-helmdocs.sh` to generate the templates? We might want to keep using this wrapper incase things change.

angelnu:
@onedr0p - we could do it but currently the script is ignoring the .helmdocsignore so it adds the templates to all folders.

Once the other PR is merged that .helmdocsignore will not be needed so maybe we merge this PR as it is and once the other PR is merged (including the 8 charts I did not convert yet) we can adopt `hack/gen-helmdocs.sh`

onedr0p:
ok no problem! 

What do you think about if it were to run on a cron instead of when pushes are made?

angelnu:
Yes, we can transform to use an scheduled trigger since we can very quickly process all the charts anyway.

Should we also consider calling in the release CI to ensure we do not build charts with outdated info? Potentially, we could also commit the change so we do not have any PRs to merge.

onedr0p:
We have to wait until we fix the chart releaser action so that it doesn't error out on duplicate tags.

It ties in with my comment on Discord

https://discord.com/channels/673534664354430999/740734177279868949/818477961190965258

angelnu:
Changed to use a daily schedule as trigger.

.github/workflows/charts-helm-docs.yaml
```diff
@@ -0,0 +1,48 @@
...
+      run: |
+        ./hack/gen-helm-docs.sh
+        
+    - name: Create pull request for helm-docs
+      uses: peter-evans/create-pull-request@v3
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
+        branch: "helm-docs/ci"
+        delete-branch: true
+        title: "chore(update helm-docs) [ci-skip]"
```
onedr0p, line 41:
```suggestion
        title: "chore(docs) update helm-docs [ci-skip]"
```

.github/workflows/charts-helm-docs.yaml
```diff
@@ -0,0 +1,48 @@
...
+        signoff: true
+        committer: "${{ github.actor }} <${{ github.actor }}@users.noreply.github.com>"
+        author: "${{ github.actor }} <${{ github.actor }}@users.noreply.github.com>"
+        commit-message: "chore(update helm-docs) [ci-skip]"
```
onedr0p, line 45:
```suggestion
        commit-message: "chore(docs) update helm-docs [ci-skip]"
```

onedr0p (changes requested):

onedr0p (approved):

PR #646, opened 2021-03-07
[helm-docs] Use helm docs for all charts [ci-skip]

angelnu:
**Description of the change**

Use helm-docs and templates for charts READMEs.

**Benefits**

Less boiler-plate for charts and ensure changelog and value descriptions are updated.

**Possible drawbacks**

**Applicable issues**

- NA

**Additional information**

**Checklist** 


angelnu:
There are 8 charts I did not migrate yet since they require more time to move the value comments from the readme into the value.yaml.

I will do those in a separate PR next weekend if nobody do them before

charts/adguard-home/README.md.gotmpl
```diff
@@ -0,0 +1,145 @@
...
+
+{{ template "chart.valuesTable" . }}
+{{- end -}}
+
+{{- define "custom.support" -}}
+## Support
+
+- See the [Wiki](https://github.com/k8s-at-home/charts/wiki)
+- Open a [issue](https://github.com/k8s-at-home/charts/issues/new/choose)
+- Ask a [question](https://github.com/k8s-at-home/charts/discussions)
```
bjw-s, line 105:
```suggestion
- See the [Docs](https://docs.k8s-at-home.com/our-helm-charts/getting-started/)
- Open an [issue](https://github.com/k8s-at-home/charts/issues/new/choose)
- Ask a [question](https://github.com/k8s-at-home/organization/discussions)
```

I think we better update these URL's right now, otherwise we'll need to do another version bump shortly to update these anyway. I will create a PR to cover the other charts if required.

bjw-s (changes requested):
I think it looks good for the most part. Thanks a lot for this massive undertaking! :-) Only a small change requested to some URL's

charts/adguard-home/README.md.gotmpl
angelnu: Ok, I wil do a full search&replace for these three strings.

angelnu:
ok, big change done - let us see how many charts require an extra bump...

If you have the time please do the PR for the remaining 8. If do not think I will have the time prior to the weekend

onedr0p (approved):

PR #650, opened 2021-03-10
[joplin-server] new chart

auricom:
[Joplin-server](url) chart based on the dependent 'common' chart.
Used to sync multiple joplin clients.

**Checklist** 


bjw-s:
It looks like you are missing the README helm-doc template files that have been added recently. Please make sure you add these as well. That way the README should be automatically generated. 🙂

auricom:
Ok, I think I understand. Nice feature !

angelnu:
@auricom - if you enable the workflow action in your fork it will check that the helm doc is also updated in the same way as we do it in the main repository.

auricom:
Thanks for helping me with this. I added the generated README.md

charts/joplin-server/Chart.yaml
```diff
@@ -0,0 +1,21 @@
...
+  - https://github.com/laurent22/joplin/tree/dev/packages/server
+  - https://github.com/laurent22/joplin
+  - https://hub.docker.com/r/joplin/server
+maintainers:
+  - name: auricom
+    email: k8s-at-home@xpander.eml.cc
+dependencies:
+  - name: common
+    repository: https://k8s-at-home.com/charts/
+    version: 2.2.0
```
angelnu, line 21: could you use the latest (3.1.0)?

charts/joplin-server/values.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+strategy:
+  type: Recreate
+
+service:
+  port:
+    port: 22300
+
+env:
+  # APP_BASE_URL: https://joplin.domain
```
angelnu, line 16: Could you annotate this values? 
https://github.com/norwoodj/helm-docs

charts/joplin-server/Chart.yaml
angelnu: In the meanwhile the common has moved and should look like:
```
- name: common
  repository: https://library-charts.k8s-at-home.com
  version: 1.0.0
```

onedr0p:
@all-contributors please add @auricom for code

charts/joplin-server/values.yaml
auricom: Ok, I added annotation like you did on [librespeed chart](https://github.com/angelnu/charts/blob/master/charts/librespeed/values.yaml)

PR #664, opened 2021-03-13
[calibre] new chart

rolfberkenbosch:
**Description of the change**
Calibre chart with webbased acces to the desktop app. Very handy!

**Checklist** 

### DIFF
--- a/charts/calibre/Chart.yaml
+++ b/charts/calibre/Chart.yaml
@@ -1,10 +1,10 @@
 ---
 
 apiVersion: v2
-appVersion: 1.0.0
+appVersion: 5.13.0
 description: Calibre is a powerful and easy to use e-book manager.
 name: calibre
-version: 5.13.0
+version: 1.0.0
 kubeVersion: ">=1.16.0-0"
 keywords:
   - calibre
--- a/charts/calibre/Chart.yaml
+++ b/charts/calibre/Chart.yaml
@@ -1,0 +1,22 @@
+---
+
+apiVersion: v2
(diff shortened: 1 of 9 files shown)


charts/calibre/Chart.yaml
```diff
@@ -0,0 +1,22 @@
+---
+
+apiVersion: v2
+appVersion: 1.0.0
```
onedr0p, line 4: This needs to be 5.13.0

charts/calibre/Chart.yaml
```diff
@@ -0,0 +1,22 @@
...
+description: Calibre is a powerful and easy to use e-book manager.
+name: calibre
+version: 5.13.0
```
onedr0p, line 7: This needs to be 1.0.0

onedr0p:
Just noticed this the version mistake on wikijs too, I opened a PR for that here https://github.com/k8s-at-home/charts/pull/666

I had to bump the version to 3.0.0 because we cannot "downgrade" a version.

charts/calibre/Chart.yaml
rolfberkenbosch: Ik fix the problem.

charts/calibre/Chart.yaml
rolfberkenbosch: Ik fix the problem.

onedr0p (approved):

onedr0p:
@all-contributors please add @rolfberkenbosch for code

PR #682, opened 2021-03-16
Emby

acjohnson:
**Description of the change**

Added Emby helm chart based on existing Plex helm chart

**Benefits**

Allows home k8s users to deploy Emby

**Possible drawbacks**

I'm not thrilled with the way I added `xmlstarlet` to the embyserver container on init, this should be part of the image or if a custom image is undesirable then maybe it should just be replaced with some simple `echo` commands...


onedr0p:
@acjohnson was there a reason you choose not to use our common library? Have you seen the jellyfin chart which uses it?

onedr0p:
Also worth noting, that with Jellyfin and Plex, we needed to implement a `podSecurityContext` in-order for hardware acceleration to work. See this https://github.com/k8s-at-home/charts/pull/684

acjohnson:
I'll have to take a closer look, I think what made me avoid the common library is the bad links in the README.md files in the various chart repos,  I guess I wasn't sure what was going on with it

https://github.com/k8s-at-home/charts/blame/master/charts/plex-media-server/README.md#L79
https://github.com/k8s-at-home/charts/blame/master/charts/plex-media-server/README.md#L99
https://github.com/k8s-at-home/charts/blame/master/charts/jellyfin/README.md#L75
https://github.com/k8s-at-home/charts/blame/master/charts/jellyfin/README.md#L93

The do not use messages might have scared me off as well :smiley: 

https://github.com/k8s-at-home/charts/blame/master/charts/plex-media-server/README.md#L71

onedr0p:
The common library is nothing to be too afraid of, in most cases it simplifies a lot of config.

acjohnson:
Awesome, sorry for the dumb question but where do I find the common library?

onedr0p:
See the bottom of this doc: https://docs.k8s-at-home.com/our-helm-charts/common-library/

But might we worth to look over the other pages there :)

acjohnson:
Welp I thought I understood helm chart writing until I learned there is such a thing as library-charts :sweat_smile:

not sure when I'll get to it, I suspect it will be a significant rewrite of this chart and might require changes to the common library chart as well.

onedr0p:
> Welp I thought I understood helm chart writing until I learned there is such a thing as library-charts :sweat_smile:
> 
> not sure when I'll get to it, I suspect it will be a significant rewrite of this chart and might require changes to the common library chart as well.

You could easily create it like the plex-media-server chart without much extra configuration.

PR #695, opened 2021-03-19
[deconz] Migrate to common (reason: Fix labels that prevent upgrading)

CuBiC3D:
**Description of the change**

Removed version labels (change every release) from `spec.VolumeClaimTemplate.metadata.labels` of the `StatefulSet`. Those template fields are immutable on `StatefulSets` and do prevent any upgrades to the chart by Helm.

```
Error: UPGRADE FAILED: cannot patch "deconz" with kind StatefulSet: StatefulSet.apps "deconz" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden
```
See https://github.com/helm/charts/issues/7803

Also adds back the `timeoutSeconds: 10` value for readiness probe - the documenting comment was there but the value somehow disappeared. Empty fields that will be populated with standard values by kubernetes API can lead problems with tools like ArgoCD that check for resource diversion.

**Benefits**
Upgrading possible ;)

**Possible drawbacks**
Breaking change will require to delete the old `StatefulSet` or uninstall release. But upgrades currently not work anyways.

**Applicable issues**

none

**Additional information**

**Checklist** 


onedr0p:
I wonder if it's worth coverting this to use the common library. 🤔

CuBiC3D:
> I wonder if it's worth coverting this to use the common library.

Already thought about that and it probably is, but should we do that prior to fix or fix and convert after?

bjw-s:
I would prefer doing it now... That way we only have one breaking change, and the labels problem will be fixed "automatically" as well.

CuBiC3D:
Then let's close this and I'll move it over to common today/weekend ;)

CuBiC3D:
Actually we can keep this branch, I'll change it to draft for now

CuBiC3D:
Migrated to common chart and documented necessary configuration with comments in `values.yaml`. Small migration info is in readme. Upgraded on real cluster+device with `helm update`, runs fine. Let me know if there's something to improve :)

CuBiC3D:
I'll take a look at CI later ;)

onedr0p (approved):

PR #698, opened 2021-03-20
[plex] migrate to common library

onedr0p:
**Description of the change**

- Migrate Plex to the common library and use our k8s@home Plex container image.
- Deprecate the old `plex-media-server` chart

**Benefits**

Easier to maintainer moving forward

**Possible drawbacks**

No clear migration path (as stated in the changelog)

**Applicable issues**

- fixes #

**Additional information**

**Checklist** 


charts/stable/plex/values.yaml
```diff
@@ -1,371 +1,112 @@
...
-# The Image to use for PLEX
+#
+# IMPORTANT NOTE
+#
+# This chart inherits from our common library chart. You can check the default values/options here:
+# https://github.com/k8s-at-home/library-charts/tree/main/charts/stable/common/values.yaml
+#
 
 image:
-  repository: plexinc/pms-docker
```
billimek, line 8: Are we sure we want to move away from the 'official' plex image here?

onedr0p: I am! People can use to use the upstream image but will lose out on some custom configuration. Someday we will update the other charts to use our images. We're getting better at documention :)

billimek (approved):
LGTM

PR #705, opened 2021-03-22
[unifi] Updated AppVersion to 6.0.45

spyder007:
**Description of the change**

Modified the Ingress definitions to use the latest Ingress spec (networking.k8s.io/v1) for clusters 1.18+ and the beta spec (networking.k8s.io/v1beta1) 1.14+.  Otherwise, continue using extensions/v1beta1.

**Benefits**

Latest Unifi controller and support for current Kubernetes versions, as well as updated template to generate labels in _helpers for consistency

**Possible drawbacks**

I do not have a version of Kubernetes running previous to v1.20, so my testing of the Ingress changes was limited to just that case.  A quick check of the chart on older kubernetes versions would be helpful, although, I'm not sure what you're policies are for support of previous kubernetes versions

**Applicable issues**

None

**Additional information**

**Checklist** 
- [ ] (optional) Chart is using our [common](https://github.com/k8s-at-home/charts/tree/master/charts/common) library as a dependency. *No Dependency Changes*
- [] (optional) Variables are documented in the README.md - *N/A - No Variable Changes*


bjw-s (commented):
I think the change looks good in general, thanks for writing this! 👍🏻

We do something similar in our common library chart:
https://github.com/k8s-at-home/library-charts/blob/5a5a6167702ba247a06615a5185864ae11739852/charts/stable/common/templates/lib/chart/_capabilities.tpl#L71
https://github.com/k8s-at-home/library-charts/blob/main/charts/stable/common/templates/classes/_ingress.tpl

spyder007:
I updated the logic to be closer to what you have in the common library template.

carpenike:
@bjw-s -- the update look good to you?

bjw-s:
> @bjw-s -- the update look good to you?

Yes, /lgtm!

bjw-s (approved):

PR #714, opened 2021-03-25
[adguard-home] fix  volumes subpath

abualy:
**Description of the change**

the goal of the change is to : 
- reflect subpath config in the values.yaml to show in the deployment

**Benefits**

- actually use subpath in volume mounts in the deployment when a user sets this parameter and expects it to work

**Possible drawbacks**

should be a seamless change for existing users, so can't think of any

**Applicable issues**

None

**Checklist** 

### DIFF
--- a/charts/stable/adguard-home/templates/deployment.yaml
+++ b/charts/stable/adguard-home/templates/deployment.yaml
@@ -65,7 +65,7 @@ spec:
               readOnly: false
             - name: config
               mountPath: /opt/adguardhome/conf
-            {{- with .Values.persistence.conf.subPath }}
+            {{- with .Values.persistence.config.subPath }}
               subPath: {{ . }}
             {{- end }}
               readOnly: false
--- a/charts/stable/adguard-home/templates/deployment.yaml
+++ b/charts/stable/adguard-home/templates/deployment.yaml
@@ -59,9 +59,11 @@ spec:
           volumeMounts:
             - name: work
               mountPath: /opt/adguardhome/work
+              subPath: {{ .Values.persistence.work.subPath }}
               readOnly: false
             - name: config
               mountPath: /opt/adguardhome/conf
+              subPath: {{ .Values.persistence.conf.subPath }}
               readOnly: false
         {{- end }}
       {{- end }}
@@ -79,9 +81,11 @@ spec:
           volumeMounts:
             - name: work
               mountPath: /opt/adguardhome/work
+              subPath: {{ .Values.persistence.work.subPath }}
               readOnly: false
             - name: config
               mountPath: /opt/adguardhome/conf
+              subPath: {{ .Values.persistence.conf.subPath }}
               readOnly: false
             {{- if .Values.tlsSecretName }}
             - name: certs
--- a/charts/stable/adguard-home/templates/deployment.yaml
+++ b/charts/stable/adguard-home/templates/deployment.yaml
@@ -65,8 +65,8 @@ spec:
               readOnly: false
             - name: config
               mountPath: /opt/adguardhome/conf
-            {{- if .Values.persistence.conf.subPath }}
-              subPath: {{ .Values.persistence.conf.subPath }}
+            {{- with .Values.persistence.conf.subPath }}
+              subPath: {{ . }}
             {{- end }}
               readOnly: false
         {{- end }}
--- a/charts/stable/adguard-home/templates/deployment.yaml
+++ b/charts/stable/adguard-home/templates/deployment.yaml
@@ -59,11 +59,15 @@ spec:
           volumeMounts:
             - name: work
               mountPath: /opt/adguardhome/work
+            {{- if .Values.persistence.work.subPath }}
               subPath: {{ .Values.persistence.work.subPath }}
+            {{- end }}
               readOnly: false
             - name: config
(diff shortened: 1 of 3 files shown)


bjw-s (commented):
Thanks a lot for taking the time to create this PR! 👍 
I have some reservations about the Service change, but since I'm not actually using the chart I will defer to the other maintainers. (I think @carpenike is actively running it?)

charts/stable/adguard-home/templates/service.yaml
```diff
@@ -26,7 +26,7 @@ spec:
   {{- end }}
   externalTrafficPolicy: {{ .Values.service.externalTrafficPolicy }}
   ports:
-    - port: 3000
+    - port: 80
```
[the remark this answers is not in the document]
abualy: tbh, thinking about it, i agree, might as well keep 3000, so not to add to confusion, or match the port being used by the pods. but that could be in a separate PR
so for now i will change PR to handle only the `subPath` bug
thank you @bjw-s

charts/stable/adguard-home/templates/deployment.yaml
```diff
@@ -59,9 +59,15 @@ spec:
           volumeMounts:
             - name: work
               mountPath: /opt/adguardhome/work
+            {{- if .Values.persistence.work.subPath }}
+              subPath: {{ .Values.persistence.work.subPath }}
+            {{- end }}
```
bjw-s, line 64:
```suggestion
            {{- with .Values.persistence.work.subPath }}
              subPath: {{ . }}
            {{- end }}
```

charts/stable/adguard-home/templates/deployment.yaml
```diff
@@ -59,9 +59,15 @@ spec:
...
               readOnly: false
             - name: config
               mountPath: /opt/adguardhome/conf
+            {{- if .Values.persistence.conf.subPath }}
+              subPath: {{ .Values.persistence.conf.subPath }}
+            {{- end }}
```
bjw-s, line 70:
```suggestion
            {{- with .Values.persistence.conf.subPath }}
              subPath: {{ . }}
            {{- end }}
```

charts/stable/adguard-home/templates/deployment.yaml
```diff
@@ -79,9 +85,15 @@ spec:
           volumeMounts:
             - name: work
               mountPath: /opt/adguardhome/work
+            {{- if .Values.persistence.work.subPath }}
+              subPath: {{ .Values.persistence.work.subPath }}
+            {{- end }}
```
bjw-s, line 90:
```suggestion
            {{- with .Values.persistence.work.subPath }}
              subPath: {{ . }}
            {{- end }}
```

charts/stable/adguard-home/templates/deployment.yaml
```diff
@@ -79,9 +85,15 @@ spec:
...
               readOnly: false
             - name: config
               mountPath: /opt/adguardhome/conf
+            {{- if .Values.persistence.conf.subPath }}
+              subPath: {{ .Values.persistence.conf.subPath }}
+            {{- end }}
```
bjw-s, line 96:
```suggestion
            {{- with .Values.persistence.conf.subPath }}
              subPath: {{ . }}
            {{- end }}
```

bjw-s (changes requested):
Looks like the CI is failing because `subPath` is optional. I added some suggestions that _should_ fix it.

abualy:
thank you @bjw-s , i have very limited experience with helm templating, so this a good opportunity to learn :D

abualy:
> Looks like the CI is failing because `subPath` is optional. I added some suggestions that _should_ fix it.

it seems weird, because i even tried setting supath parameter to default value:`""`, and not use conditional in the template, and still it gives the same linting error

charts/stable/adguard-home/templates/deployment.yaml
```diff
@@ -59,9 +59,15 @@ spec:
           volumeMounts:
             - name: work
               mountPath: /opt/adguardhome/work
+            {{- with .Values.persistence.work.subPath }}
+              subPath: {{ . }}
+            {{- end }}
               readOnly: false
             - name: config
               mountPath: /opt/adguardhome/conf
+            {{- with .Values.persistence.conf.subPath }}
```
bjw-s, line 68:
```suggestion
            {{- with .Values.persistence.config.subPath }}
```

charts/stable/adguard-home/templates/deployment.yaml
```diff
@@ -79,9 +85,15 @@ spec:
...
```
bjw-s, line 94:
```suggestion
            {{- with .Values.persistence.config.subPath }}
```

bjw-s (changes requested):
The item is called `persistence.config`, not `persistence.conf`, so it couldn't find that value when rendering the template. I missed that in my initial code suggestion, hopefully this fixes that 🙂

charts/stable/adguard-home/templates/deployment.yaml
abualy: :facepalm:

bjw-s (approved):

PR #719, opened 2021-03-25
[ser2sock] migrate to common library

onedr0p:
**Description of the change**

Migrate ser2sock to the common library

**Checklist** 


onedr0p:
This chart is one we exclude, from testing because it requires a mounted device to pass tests. @billimek I don't use this, just going thru updating some charts to common. Let me know what you think!

billimek:
It's really too bad there isn't a way to run this such that it fails if there isn't a physical device present.  Long ago I looked for an alternative container that does a similar thing and came up empty.  This does serve a useful purpose too.

charts/stable/ser2sock/values.yaml
```diff
@@ -1,73 +1,55 @@
...
-nodeSelector: {}
+strategy:
+  type: Recreate
 
-# -- Toleration labels for pod assignment of the GUI
-tolerations: []
+env:
+  # TZ:
+  # PUID:
+  # PGID:
```
billimek, line 19: I'm OK with leaving these unset in the new version, but we had these specifically set before in the values to default to  `1001`.

Do we want to keep that default behavior, or leave it undefined, as a best practice?

onedr0p: Usually we've been keeping them commented out as it really depends on the peoples setup.

billimek (dismissed):
LGTM!

onedr0p:
@bjw feel free to merge unless you want to test this in the CI

bjw-s:
Yes, I'll use this PR to fix that CI thingy, will merge afterwards 👍

PR #725, opened 2021-03-26
[digitalocean-dyndns/ddclient] deprecate charts

onedr0p:
**Description of the change**

deprecate both digitalocean-dyndns and ddclient charts

**Benefits**

It's a much more elegant solution to use a k8s CronJob to do this, rather then support charts for any providers dyndns chart.

I have laid out a guide below on how to use a configmap, secret and cronjob to accomplish this task. We should point people to it when asked or issues are raised relating on how to handle dyndns in your cluster

https://docs.k8s-at-home.com/guides/dyndns/

**Additional information**

We could perhaps support a chart that wires this up automagically, but it might be hard because of all the different DNS APIs

**Checklist** 

### DIFF
--- a/charts/stable/ddclient/Chart.yaml
+++ b/charts/stable/ddclient/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 3.9.1
 description: Perl client used to update dynamic DNS entries for accounts on Dynamic DNS Network Service Providers
 name: ddclient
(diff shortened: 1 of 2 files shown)


onedr0p:
We could perhaps support a chart that wires this up automagically, but it might be hard because of all the different DNS APIs

bjw-s (approved):

PR #728, opened 2021-03-26
[valheim] - set realistic resource defaults for valheim

billimek:
**Description of the change**

The default resource settings from the common lib chart aren't adequate enough for the resources needed for the valheim container.  This change sets reasonable CPU & memory requests/limits for running this particular workload.

**Benefits**

This will prevent OOMkills of valheim when deploying the chart

**Possible drawbacks**

Kubernetes will not be able to schedule this chart if there are not enough resources.

**Additional information**

Data collected from Prometheus about the actual CPU & memory usage of the valheim chart supporting this change:
image

**Checklist** 

### DIFF
--- a/charts/stable/valheim/Chart.yaml
+++ b/charts/stable/valheim/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: latest
 description: Valheim dedicated gameserver with automatic update and world backup support
 name: valheim
-version: 2.0.1
+version: 2.0.2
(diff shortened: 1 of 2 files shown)


onedr0p (approved):

bjw-s:
This is interesting, since we don't set any default resources in the library. It shouldn't OOMKill then if I understand correctly:

https://github.com/k8s-at-home/library-charts/blob/0a91f09204d7f8d152b55159296246dcd9ac78d2/charts/stable/common/values.yaml#L282

PR #739, opened 2021-03-31
[postgres-cluster] New chart to create Postgres clusters with local storage

angelnu:
**Description of the change**

This chart is a wrapper for the [Zalando postgres operator](https://github.com/zalando/postgres-operator) to create
a high available Postgres cluster using nodes local storage.

**Benefits**

Features added by this wrapper:
- creates an storage class using local storage in the nodes specified in `persistentVolumes.replicaNodes`. These replicas survive
  a tear-down and tear-up of the cluster.
- (optional) does sql dumps to an existing PVC with K8S cronjobs. The Operator only supports backups to cloud, not NFS PVC for example.
- Define the password for the DB so that it can be used accross tear-down and tear-up of the cluster.

**Possible drawbacks**

Another chart to maintain - I can offer the maintenance.

**Applicable issues**

- fixes #

**Additional information**

Since it depends on the operator being loaded previously and also knowing the topology of the cluster I have disabled the install test and only do lint.

**Checklist** 


charts/stable/postgres-cluster/templates/storageClass.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: {{ include "postgres-cluster.storageClass" . }}
+  labels:
+    {{- include "common.labels" . | nindent 4 }}
+  annotations:
+provisioner: kubernetes.io/no-provisioner
+reclaimPolicy: Delete
+volumeBindingMode: WaitForFirstConsumer
```
onedr0p, line 10: New at the end of the file here :)

onedr0p (changes requested):
This looks like an interesting chart! Just one small change and I'll trust you to help maintain it :D

onedr0p:
Do you think this should be called something like `zalando-postgres-cluster`?

angelnu:
Yes, I can rename it - will do now together with the missing end of file new line.

Sure I will maintain it - I have 6 postgres clusters myself using this! I find easier to deal with DB issues and going back when the DB is used only by a single chart.

@onedr0p - if we consider this chart as a dependency for other charts in this repo - do we still it put it here or under the library repository?

onedr0p:
I think leaving it here is fine, we can always change it in the future if we feel differently 🙂

onedr0p (approved):

PR #741, opened 2021-03-31
[adguard-home] migrate to common

onedr0p:
**Description of the change**

Migrate AGH to the common library

With the help of an `initContainer` the `configmap` is copied into `/opt/adguardhome/work/AdGuardHome.yaml` if it does not exist. The reason for this is AGH will not be able to write to this file on AGH updates or UI config changes if we simply reference the `configmap` because it is immutable. This way you can set a config persistence to keep the file state across pod restarts.

If people want to keep all config as code they should update the `args` to be something like

```yaml
args:
- "--config /tmp/AdGuardHome.yaml"
```

This will force AGH to use the `configmap` as config and users can disable persistence on the config volume but then folks won't be able to make any changes on the UI or upgrade AGH due to file migrations.

**Additional information**

**Checklist** 


charts/stable/adguard-home/README.md
```diff
@@ -72,187 +75,45 @@ N/A
...
...tps://adguardteam.github.io/AdGuardSDNSFilter/Filters/filter.txt\n  name: AdGuard DNS filter\n  id: 1\n- enabled: false\n  url: https://adaway.org/hosts.txt\n  name: AdAway\n  id: 2\n- enabled: false\n  url: https://www.malwaredomainlist.com/hostslist/hosts.txt\n  name: MalwareDomainList.com Hosts List\n  id: 4\nwhitelist_filters: []\n# - enabled: true\n#   url: https://easylist-downloads.adblockplus.org/exceptionrules.txt\n#   name: Allow nonintrusive advertising\n#   id: 1595760241\nuser_rules: []\n# - '||example.org^'\n# - '@@||example.org^'\n# - 127.0.0.1 example.org\n# - '! Here goes a comment'\n# - '# Also a comment'\ndhcp:\n  enabled: false\n  interface_name: \"\"\n  gateway_ip: \"\"\n  subnet_mask: \"\"\n  range_start: \"\"\n  range_end: \"\"\n  lease_duration: 86400\n  icmp_timeout_msec: 1000\nclients: []\n# - name: myuser\n#   tags:\n#   - user_admin\n#   ids:\n#   - 192.168.91.1\n#   use_global_settings: true\n#   filtering_enabled: false\n#   parental_enabled: false\n#   safesearch_enabled: false\n#   safebrowsing_enabled: false\n#   use_global_blocked_services: true\n#   blocked_services: []\n#   upstreams: []\nlog_file: \"\"\nverbose: false\nschema_version: 6\n"` |  |
```
bjw-s, line 78: We really need to find some way to fix this :/

onedr0p: I can fix it here I think

onedr0p: Oh wait... no, it's the same as before :(

charts/stable/adguard-home/README_CHANGELOG.md.gotmpl
```diff
@@ -9,19 +9,19 @@ All notable changes to this application Helm chart will be documented in this fi
...
 
-- Use helm-docs
+- **BREAKING** Migrate Adguard Home to the common library, a lot of configuration has changed.
 
 #### Removed
 
 - N/A
 
-[2.2.1]: #2.2.1
+[6.0.0]: #6.0.0
```
bjw-s, line 26:
```suggestion
[3.0.0]: #3.0.0
```

bjw-s (changes requested):
Just a small nitpick as far as I can tell

bjw-s (approved):

PR #749, opened 2021-04-01
[zwavejs2mqtt] Update image and add websocket port example

Ardetus:
**Description of the change**

**Benefits**

**Possible drawbacks**

**Applicable issues**

- fixes #

**Additional information**

**Checklist** 


charts/stable/zwavejs2mqtt/values.yaml
```diff
@@ -21,6 +21,9 @@ env: {}
 service:
   port:
     port: 8091
+  # additionalPorts:
+  # ┊ - name: websocket
+  # ┊   port: 3000
```
onedr0p, line 26: Are you sure you don't want to add an `additionalService` instead, would a `additionalPort` be fine for this usecase?

onedr0p (commented):
Which one should we use here.

charts/stable/zwavejs2mqtt/values.yaml
Ardetus: Not really sure what is proper way... I copied this from deconz chart and it worked fine on my setup

onedr0p: Sounds good, we can roll with it :)

onedr0p (approved):

onedr0p:
@all-contributors add @Ardetus for code

PR #774, opened 2021-04-09
statping - Fix #747

### LINKED ISSUE #747: [statping] hardcoded env causes undesired results
# Details

statping-3.1.1

**Container name and tag:**

N/A

Defined some values via `env:` and some via `envFrom:`

**Relevant Helm values:**

```yaml
  env:
    NAME: cajun.pro status
    DESCRIPTION: cajun.pro hosting status page
    VIRTUAL_HOST: https://status.cajun.pro
    DB_CONN: postgres
    DB_HOST: postgres-statping
    DB_DATABASE: postgres
  envFrom:
    - secretRef:
        name: statping
```
https://github.com/DirtyCajunRice/GitOps-Argo-CD/blob/2ebaec880e25deee3cdba1aa09a44e7237f11449/statping/values.yaml

for only the values i defined in env to be used

N/A

**Additional Information:**

Adding a check for `ENVVARNAME: "-"` is going to be too verbose. Some type of globally set var should toggle using all defaults or none. There are workarounds for wiping the env dict like `env: ~` but that doesnt work when the chart checks and sets anyway.

DirtyCajunRice:
**Description of the change**

stops importing postgresql env vars when the chart is not enabled and allows disabling of the admin env vars

**Benefits**

allows better use of the common chart functions

**Possible drawbacks**

n/a

**Applicable issues**

- fixes #747 

**Additional information**

**Checklist** 


bjw-s:
Could it be easier / cleaner to comment out the env vars in the values.yaml instead of adding an additional boolean?

DirtyCajunRice:
@bjw-s  Ok. Modified. If you agree with https://github.com/k8s-at-home/library-charts/issues/19 this will be short lived, but will fix now as 19 is a larger task to find and replace everyhwere.

charts/stable/statping/values.yaml
```diff
@@ -8,7 +8,9 @@
 image:
   repository: statping/statping
   pullPolicy: IfNotPresent
-  tag: "v0.90.65"
+  # -- image tag
+  # @default -- `.Chart.appVersion`
+  tag: ""
```
bjw-s, line 13: Even though the common library will default to the appVersion, we still keep an explicit tag in our values.yaml, also so that it gets populated in the README.

bjw-s (changes requested):
Small change requested, rest of the changes look good 👍

charts/stable/statping/values.yaml
DirtyCajunRice: updated

PR #777, opened 2021-04-11
[reg] Add reg chart

nicholaswilde:
Signed-off-by: Nicholas Wilde <ncwilde43@gmail.com>

**Description of the change**

Adds `reg` chart

**Benefits**

New chart.

**Possible drawbacks**

None

**Applicable issues**

None

**Additional information**

`env.EXTRA_ARGS` is commented out in `values.yaml` in order to give direction on how to start the reg server.

**Checklist** 


charts/stable/reg/README_CHANGELOG.md.gotmpl
```diff
@@ -0,0 +1,27 @@
...
+
+- N/A
+
+#### Changed
+
+- N/A
+
+#### Removed
+
+- N/A
```
bjw-s, line 24:
```suggestion
#### Added

- Initial version
```

charts/stable/reg/Chart.yaml
```diff
@@ -0,0 +1,20 @@
...
+apiVersion: v2
+appVersion: 0.16.1
+description: Docker registry v2 command line client and repo listing generator with security checks.
+name: reg
+version: 1.0.0
+kubeVersion: ">=1.16.0-0"
+keywords:
+- reg
+home: https://github.com/k8s-at-home/charts/tree/master/charts/stable/reg
+sources:
```
bjw-s, line 11: It's normally a good practice to also include some sort of icon. Maybe there is some sort of Github avatar for `genuinetools` that can be "borrowed"?

charts/stable/reg/values.yaml
```diff
@@ -0,0 +1,27 @@
...
+  pullPolicy: IfNotPresent
+  tag: v0.16.1
+
+strategy:
+  type: Recreate
+
+# See more environment variables in the reg documentation
+# https://github.com/genuinetools/reg
+env:
+  EXTRA_ARGS: "server --registry r.j3ss.co"
```
bjw-s, line 19: I'm not familiar with the app, but is this a good value to have as default?

bjw-s (changes requested):

charts/stable/reg/Chart.yaml
nicholaswilde: I can take a look. I know that there isn't one for the app.

charts/stable/reg/values.yaml
nicholaswilde: The ci tests fail if we don't specify a server to run.

bjw-s: Ah, in that case you can create a `ci/ct-values.yaml` file containing this. That way the CI will use that and we can remove the default here.

nicholaswilde: @bjw-s I'm trying to determine the best route to implement `ci/ct-values.yaml`. I'm looking at `charts-lint-test.yaml` and trying to see if I should add on to the `changes-install` and `install` jobs in `charts-lint-test.yaml`, or if I should just create a whole new job. Which would you recommend?

onedr0p: You just need to place values in `ci/ct-values.yaml` that you want to run only on CI

bjw-s (approved):
I took the liberty of adding the ci/ct-values.yaml file, because the chart looks good otherwise. Merging! 🎉

nicholaswilde:
> I took the liberty of adding the ci/ct-values.yaml file, because the chart looks good otherwise. Merging! 🎉 

I appreciate it! Thanks for the help!

PR #778, opened 2021-04-11
[adguard-home] fix conf mount path

n0rad:
The init container do not share the conf directory adguard container. So Moving the conf doing nothing.

This PR make the conf directory shared between the 2 containers and so fix as code configuration

### DIFF
--- a/charts/stable/adguard-home/Chart.yaml
+++ b/charts/stable/adguard-home/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: v0.105.2
 description: DNS proxy as ad-blocker for local network
 name: adguard-home
-version: 3.1.0
+version: 3.1.1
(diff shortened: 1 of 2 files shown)


bjw-s:
Good catch on this one! Your modification to the initContainer should work, but I don't think the additionalVolumes changes will work. Would it be OK if I submit some changes to your PR to see if I can fix it / make the initContainer depend on the existing persistence settings?

n0rad:
Sure ! fill free to commit, I was not sure on how to fix the conflict between the user defined volume and the empty dir volume needed to do the copy.

onedr0p:
@n0rad I think this might be a good workaround

https://github.com/k8s-at-home/charts/pull/785

onedr0p:
@all-contributors add @n0rad for code

PR #780, opened 2021-04-13
[bitwardenrs] [FIX] Add SMTP password

lcotonea:
**Adding the support of SMTP password**

**Benefits**
If not set, we cannot use SMTP with authentication

**Checklist** 

### DIFF
--- a/charts/stable/bitwardenrs/templates/configmap.yaml
+++ b/charts/stable/bitwardenrs/templates/configmap.yaml
@@ -29,6 +29,9 @@ data:
   {{- if and (not .existingSecret.enabled) .user }}
   SMTP_USERNAME: {{ .user | quote }}
   {{- end }}
+  {{- if .password }}
+  SMTP_PASSWORD: {{ .password | quote }}
+  {{- end }}
   {{- end }}
   {{- end }}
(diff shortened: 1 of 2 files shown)


charts/stable/bitwardenrs/templates/configmap.yaml
```diff
@@ -29,6 +29,9 @@ data:
   {{- if and (not .existingSecret.enabled) .user }}
   SMTP_USERNAME: {{ .user | quote }}
   {{- end }}
+  {{- if .password }}
+  SMTP_PASSWORD: {{ .password | quote }}
+  {{- end }}
```
onedr0p, line 34:
```suggestion
  {{- if and (not .existingSecret.enabled) .password }}
  SMTP_PASSWORD: {{ .password | quote }}
  {{- end }}
```

Shouldn't it be this?

onedr0p (changes requested):
You need to also bump the version in Chart.yaml

charts/stable/bitwardenrs/templates/configmap.yaml
lcotonea: Yes it should be. :+1:

PR #813, opened 2021-04-21
[transmission] use string config and showoff initContainer

onedr0p:
**Description of the change**

- Switch from `transmissionConfig` to `config` and make it a string (for some consistency across charts)
- Showoff initContainer for adding a custom webUI

**Checklist** 


charts/stable/transmission/README_CHANGELOG.md.gotmpl
```diff
@@ -9,6 +9,16 @@ All notable changes to this application Helm chart will be documented in this fi
...
 
+### [5.0.0]
+
+#### Added
+
+- Example initContainer to showcase adding a custom web ui
+
+#### Changed
+
+- **Breaking**: `transmissionConfig` became `config` and is now a string not a object
```
bjw-s, line 20:
```suggestion
- **Breaking**: `transmissionConfig` became `config` and is now a string not an object
```

charts/stable/transmission/templates/configmap.yaml
```diff
@@ -6,4 +6,4 @@ metadata:
     {{- include "common.labels" . | nindent 4 }}
 data:
   settings.json: |
-    {{- .Values.transmissionConfig | mustToPrettyJson | nindent 4 }}
+{{ .Values.config | indent 4 }}
```
bjw-s, line 9:
```suggestion
    {{- .Values.config | nindent 4 }}
```

bjw-s (changes requested):

bjw-s (approved):

ishioni:
@bjw-s oh no, why did you remove the yaml to json conversion? Now one has to redefine the whole config to change just one option :(

bjw-s:
It wasn't me who did the change, I merely approved the PR 😉

This PR makes it so that more of our charts use the same approach, which we believe is ultimately a better user experience.
In most other charts where we have config files that we pass into an app, we do this through a "simple" string so that it is immediately clear what is being passed through. This doesn't mean that the original solution wasn't a cool solution, but the benefit of having a consistent approach across our charts outweighs any potential downside of modifying a JSON file vs a YAML value.

ishioni:
I'll have to politely disagree. There was a lot of stuff that we left behind when using common which took a while to course correct, and this is an objectively worse user experience, but done for the sake of consistency. You'd have a point if the settings were obscuring some underlying mechanic in this app, but they don't in this case, making it needlessly more complicated. 

Are we to stop passing in settings in yaml in apps that use yaml to begin with for consistency with those that don't, to bring them down to a common denominator? That would be foolish as well, and I don't see those changes being made to zigbee2mqtt or icbinv. Therefore i once again ask that this be reverted.

onedr0p:
I'm trying to understand the technical benefits of storing this as yaml opposed to a string. 

As for user preference, I don't think it's any harder or easier for someone consuming this chart to use a string over yaml...

At the end of the day a configmap is made and the deployment works or not.

I'm not opposed to changing it back I'm just trying to understand what is the issue with it being a string besides difference in opinions.

ishioni:
@onedr0p 
I've outlined this in my first comment. In the original, one could override just one value, like this
```yaml
config:
  rpc-password: "foobar"
```
And leave the rest to be inherited from the chart. In the new version the user has to copy the whole configuration block to change just one value

onedr0p:
I see what you mean now, however, I think it's a bad idea to rely on those default settings for app config.

Let me elaborate. We upgrade a value in the transmission settings, it gets approved and merged. The end user updates and complains this changed and is force to define it as code anyways to overwrite. 

I know you will never run into this, but there's a lot of people who just upgrade chart versions without reading what is happening.

For me, I'd know I would want the full settings defined on my end to avoid any issue in future updates of the chart, or application. It's more declarative and easier to tweak things without having to reference what's default in our transmission chart and in the transmission docs.

ishioni:
@onedr0p I see your argument, but this applies to ALL of the values of the chart, and has the same gotchas. This is why we have semantic versioning, no? To give another counterargument - should ingress-nginx (the kubernetes one, not the commercial nginx one) remove it's config stanza and replace it with a bare nginx config? In this context it doesn't make any sense

onedr0p:
> @onedr0p I see your argument, but this applies to ALL of the values of the chart, and has the same gotchas. This is why we have semantic versioning, no? 

Here we're only talking about a specific application though, something that is only valid for transmission. I'm only trying to convey how it's a bad idea to rely on these variables being a constant value. 

I'm guilty of not reading changes to charts on bugfixes or sometimes even minor. Do we really want to make a breaking changes update anytime we change the settings yaml? We should offload that to the user to decide what to do with their config in it's entirety.

> To give another counterargument - should ingress-nginx (the kubernetes one, not the commercial nginx one) remove it's config stanza and replace it with a bare nginx config? In this context it doesn't make any sense

They do things both ways as well. They have snippet block annotations were you can define raw nginx configuration because having those certain things make no sense in yaml and could possibly break on version changes.

ishioni:
>'m only trying to convey how it's a bad idea to rely on these variables being a constant value
Yet we do that for a multitude of other charts, and almost any other chart source does this as well (anything from bitnami for example). If we go to the lowest common denominator of "the user upgrades and doesn't care about changes" then there's no reason of even having a versioning system, nor any notion of backwards compatibility.

>Do we really want to make a breaking changes update anytime we change the settings yaml?
Again, we already do this, as does any other chart anywhere, where any change warrants a version bump. This is why we have a versioning system. 

>They do things both ways as well. They have snippet block annotations[..]
Yes, but they don't throw out the child with the bathwater, do they?

bjw-s:
Another argument why I think it's not a great idea in general to predefine configuration files in configMaps, is that the contents of a config file is (often) tightly coupled to the application version that is being ran. 
Since we give the option to specify image tags, a user could end up running an appVersion that might not be compatible with the predefined configMap. That would force users to wait for us to update support, but from then on the chart would only work for appVersion >= x, leaving behind people that for whatever reason want to run version x-1.

> it got broken for no good reason. 

I think we're going to have to agree to disagree.

ishioni:
>I think we're going to have to agree to disagree.

If that's the case, fair enough

sparhomenko:
Without judgement on the discussion above, can I make another observation on a consequence of this change: some settings (namely `rpc-username` and `rpc-password`) are sensitive. Before the change, it was possible to load from a Kubernetes `Secret`, for example with Flux 2:
```
apiVersion: helm.toolkit.fluxcd.io/v2beta1
kind: HelmRelease
spec:
    spec:
      sourceRef:
        kind: HelmRepository
        name: k8s-at-home
      chart: transmission
...
  valuesFrom:
  - kind: Secret
    name: transmission
    valuesKey: user
    targetPath: transmissionConfig.rpc-username
  - kind: Secret
    name: transmission
    valuesKey: password
    targetPath: transmissionConfig.rpc-password
```
Using a `Secret` was a nice way to both keep sensitive values out of version control, and to share them with the clients that connect via RPC, eg. flood:
```
apiVersion: helm.toolkit.fluxcd.io/v2beta1
kind: HelmRelease
spec:
  chart:
    spec:
      sourceRef:
        kind: HelmRepository
        name: k8s-at-home
      chart: flood
...
  valuesFrom:
  - kind: Secret
    name: transmission
    valuesKey: user
    targetPath: env.FLOOD_OPTION_TRUSER
  - kind: Secret
    name: transmission
    valuesKey: password
    targetPath: env.FLOOD_OPTION_TRPASS
```

That does not seem to be possible after this change. Mounting them as environment variables instead would be a good alternative, but k8s@home image doesn't support that (linuxserver one had `USER` and `PASS`). Can you recommend another way to segregate these sensitive values?

onedr0p:
I think we can offload this to our container image and make the config 100% env driven. I'll have something today and share the PR here.

onedr0p:
@ishioni @bjw-s @sparhomenko let me know if this works for your use cases:

https://github.com/k8s-at-home/container-images/pull/106

sparhomenko:
@onedr0p Worked perfectly for my use case, thank you!

PR #818, opened 2021-04-22
[emby] New chart

nicholaswilde:
Signed-off-by: Nicholas Wilde <ncwilde43@gmail.com>

**Description of the change** Emby

**Benefits**

New chart

**Possible drawbacks**

None

**Applicable issues**

None

**Additional information**

**Checklist** 


onedr0p:
Incubator is meant for long lived testing of a chart with pretty non-standard config. This looks as if it's fine being release as stable.

charts/stable/emby/values.yaml
```diff
@@ -0,0 +1,30 @@
...
+service:
+  port:
+    port: 8096
+  additionalPorts:
+    - port: 8920
+      name: https
+      protocol: TCP
+
+ingress:
+  enabled: false
```
onedr0p, line 30: You'll probably want to include some persistence below here for config and media.

onedr0p (changes requested):

charts/stable/emby/values.yaml
nicholaswilde: I'm going to have to look into this a bit more. [Truecharts'](https://github.com/truecharts/apps/blob/56211222bff4333fdf827d34a88c62965777277c/.containers/apps/emby/Dockerfile) didn't have any `VOLUMES` , [LSIO](https://github.com/linuxserver/docker-emby/blob/master/README.md#usage) has `/data` but I don't see how Emby is mapped to it, and the official Emby Docker file is only calling out `/mnt/` but are [not referencing](https://github.com/MediaBrowser/Emby.Build/tree/master/docker-containers) them.

```
 volumes:
      - /path/to/programdata:/config # Configuration directory
      - /path/to/tvshows:/mnt/share1 # Media directory
      - /path/to/movies:/mnt/share2 # Media directory
```

nicholaswilde: It looks like /media already exists in the image and the storage locations are set using the emby GUI. I've also added the `/config` volume. Volume mounts seem to work without declaring `VOLUME` in the Dockerfile and so I don't think there's a need to update the image.

onedr0p (approved):

charts/stable/emby/values.yaml
onedr0p: Looks good man, thanks!

PR #825, opened 2021-04-24
docs: Document HOST_WHITELIST_ENTRIES for sabnzbd

ammmze:
**Description of the change**

Updated the list of ways around the host check in sabnzbd...we have a more convenient 3rd option in place already.

**Benefits**

Expose an easier way to configure configure the host whitelist

**Possible drawbacks**

none

**Applicable issues**

none

### DIFF
--- a/charts/stable/sabnzbd/README.md
+++ b/charts/stable/sabnzbd/README.md
@@ -1,6 +1,6 @@
 # sabnzbd
 
-![Version: 7.0.0](https://img.shields.io/badge/Version-7.0.0-informational?style=flat-square) ![AppVersion: v3.2.1](https://img.shields.io/badge/AppVersion-v3.2.1-informational?style=flat-square)
+![Version: 7.0.1](https://img.shields.io/badge/Version-7.0.1-informational?style=flat-square) ![AppVersion: v3.2.1](https://img.shields.io/badge/AppVersion-v3.2.1-informational?style=flat-square)
 
 Free and easy binary newsreader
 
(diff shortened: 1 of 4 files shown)


charts/stable/sabnzbd/README.md
```diff
@@ -70,10 +70,11 @@ helm install sabnzbd k8s-at-home/sabnzbd -f values.yaml
 
 **IMPORTANT NOTE:** when installing this chart for the first time you will get the follow message in your browser when trying to access Sabnzbd: `Access denied - Hostname verification failed: sabnzbd.org/hostname-check`
 
-You can do one of two things to solve this issue:
+You can do one of three things to solve this issue:
 
 1. Update the `sabnzbd.ini` config file to your `ingress` name and/or `loadBalancerIP` to the `host_whitelist` field and restart the pod, or
 2. Forward the service to your local machine with `kubectl port-forward service/sabnzbd -n default 8080:8080` and update the `host_whitelist` in the Sabnzbd Settings UI
+3. Set an environment variable named `HOST_WHITELIST_ENTRIES` with the value you would like placed in the `sabnzbd.ini`
```
onedr0p, line 77: This will get overwritten.

You need to update `README_CONFIG.md.gotmpl` and then run `./hack/gen-helm-docs.sh stable sabnzbd` in your shell.

With that said I'm fine with us only mentioning number 3 as the way to make it work now that were using the kah images.

onedr0p (changes requested):

charts/stable/sabnzbd/README.md
ammmze: 👍🏻  I've updated this. 

Side note, I also updated the error message about not having helm-docs in the gen-helm-docs.sh with a url to helm-docs for convenience.

bjw-s (approved):

onedr0p (approved):

charts/stable/sabnzbd/README.md
onedr0p: Looks good thanks!

PR #830, opened 2021-04-25
[transmission] use env config

onedr0p:
**Description of the change**

Goes along with https://github.com/k8s-at-home/container-images/pull/106/ discussed here https://github.com/k8s-at-home/charts/pull/813

Changes configmap config to env config

Let's wait until the container image is pushed and released before merging this.

**Additional information**

**Checklist** 


ishioni:
While I appreciate the question for input, this is as close to rediscovering the wheel as one can get. I'll be stepping back from the project and rolling out my own chart for personal use

onedr0p:
> I'll be stepping back from the project and rolling out my own chart for personal use

Sorry you feel that way. I can admit what I originally changed was a mistake but I was hoping to make up for it with this PR.

ishioni:
> > I'll be stepping back from the project and rolling out my own chart for personal use
> 
> Sorry you feel that way. I can admit what I originally changed was a mistake and I was hoping to make up for it with this PR.

That's ok, I don't feel any ill will :) That said, it seems we have different viewpoints on what the project was meant to do, and that's perfectly fine. Consider it an amicable split :)

onedr0p:
> That said, it seems we have different viewpoints on what the project was meant to do, and that's perfectly fine. Consider it an amicable split :)

Viewpoints on one chart, one configmap change, shouldn't make you ragequit contributing to these charts. I'd still like to have you around. You offer insight, and criticism not many people have the depth of knowing.

From the start we should have offloaded this configuration to the container image, it's all about iterating over design choices. I strongly feel this environment based configuration is much better than using a configmap in either way of it being a string or YAML objects.

ishioni:
>[..]shouldn't make you ragequit contributing to these charts

In order for me to ragequit, I'd have to write a salty rant first. Which I didn't. So let's leave the emotionally charged language at the door

>From the start we should have offloaded this configuration to the container image

That would require me to be the author of the image. Which I wasn't. But even baring that, I don't see the technical superiority of just passing through every option as an env instead of a config file. Maybe except for the fact that any change in these settings - which was your main argument in the original patch - would really be now obfuscated to anyone upgrading, IF they were to change. What's faster, looking at values.yaml, or looking at a tpl file inside of the source of the container? Which CD suite will show the difference in that kind of happenstance? Helm diff sure wouldn't

And here's the thing - for some time k8s@home has been plagues by changes for changes sake. Or for "consistency". At first common would just be that - a common lib for common media containers, which was fine because *arr containers really are made from the same cloth. But then one after another every app or chart was heavily pushed to migrate to it, no matter if it really fit or not. And this change, the "one configmap" is simply a change that took it too far for me.

Ornias1993:
> But then one after another every app or chart was heavily pushed to migrate to it, no matter if it really fit or not

Actually, about 90% of simple apps (the once people normally deploy at home in stock docker), need:
- A service
- A deployment
- Ingress
- Some storage

That's what the common chart (and other common-charts within other projects) is there to facilitate. Together with easier maintenance.

There is nothing "special" about transmission that  requires it to be a completely seperate chart. Besides one configmap, it's no different from: Jackett, Tautulli, Sonarr, etcetcetc.

To be clear:
I've had cases where I had to hack charts to fit into the common chart, yes.
But those are rare and definately more diffcult than "just" adding a configmap.

Ornias1993 (approved):
As the container change is already merged and functionality wise this shouldn't break much...
I think this should workout just fine :)

PR #832, opened 2021-04-25
[sharry] use a config blockk

Diaoul:
**Description of the change**
Use a config block for sharry and some clean up

**Benefits**
Fixes a security issue where using the chart would always expose admin:admin account
without the ability to disable it.

**Possible drawbacks**
User has to dig a bit in the configuration to configure authentication

**Applicable issues**
N/A

**Additional information**
I've tried to do a bit of clean up along the way, there may be a :bug: 

**Checklist** 

### DIFF
--- a/charts/stable/sharry/values.yaml
+++ b/charts/stable/sharry/values.yaml
@@ -27,7 +27,8 @@ ingress:
     # -- Full list of options https://eikek.github.io/sharry/doc/configure
 # @default -- see URL to default config
 config: |
- sharry.restserver {
+  sharry.restserver {
+    # base-url = "https://sharry.${DOMAIN}"
     backend {
       auth {
         fixed {
--- a/charts/stable/sharry/values.yaml
+++ b/charts/stable/sharry/values.yaml
@@ -16,7 +16,15 @@ service:
 args:
 - "/opt/sharry.conf"
 
-# -- Full list of options https://eikek.github.io/sharry/doc/configure
+ingress:
+  enabled: false
+  # -- Enable large file support with those annotations for nginx
+  # annotations:
+  #   nginx.ingress.kubernetes.io/client-body-buffer-size: "2048m"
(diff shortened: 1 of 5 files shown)


charts/stable/sharry/values.yaml
```diff
@@ -1,77 +1,40 @@
...
+  }
 
 ingress:
   enabled: false
   annotations:
-    # -- Nginx client Body Buffer Size
     nginx.ingress.kubernetes.io/client-body-buffer-size: "2048m"
-    # -- Nginx Proxy Body Size
     nginx.ingress.kubernetes.io/proxy-body-size: "2048m"
     nginx.ingress.kubernetes.io/proxy-buffering: "off"
```
onedr0p, line 40: We should probably comment these out, not everyone is using nginx

charts/stable/sharry/values.yaml
```diff
@@ -1,77 +1,40 @@
...
+          user = "admin"
+          password = "admin"
+          order = 10
+        }
+      }
+    }
+  }
 
 ingress:
   enabled: false
```
onedr0p, line 36: Move these two lines about the config block

onedr0p (changes requested):

charts/stable/sharry/values.yaml
Diaoul: Indeed

onedr0p:
Maybe you need more information in the config?

onedr0p:
...like 

```      
      bind {
        address = "0.0.0.0"
        port = {{- if .Values.service.port.port }}{{ .Values.service.port.port }}{{ end }}
      }```

Diaoul:
Yeah but the error message was not really helpful here. I had to guess that the reason of the crash was that the service was not reachable on the port. Is that the actual check that is made here?

onedr0p (approved):

PR #837, opened 2021-04-30
ingress hosts syntax error

fanlix:
indent error.


bjw-s (changes requested):
Thanks for taking the time to submit this PR. We would appreciate it if you would next time fill out our PR template. That would help us to better assess the scope and impact of a PR.

As for the change, I think it looks good, but the chart patch version needs to be bumped.

PR #840, opened 2021-04-30
transmission: move pulling of theme as a custom startup probe

jr0dd:
Due to limitations of network connectivity and initcontainers, I moved the pulling of flood theme to the startup probe instead. Tested working.

### DIFF
--- a/charts/stable/transmission/values.yaml
+++ b/charts/stable/transmission/values.yaml
@@ -5,20 +5,6 @@
 # https://github.com/k8s-at-home/library-charts/tree/main/charts/stable/common/values.yaml
 #
 
-# # Use an initContainer to download a custom web ui
-# initContainers:
-# - name: custom-webui
-#   image: curlimages/curl:7.76.1
-#   command:
-#   - "/bin/sh"
-#   - "-c"
-#   - "curl -fsSL -o /tmp/flood-for-transmission.tar.gz https://github.com/johman10/flood-for-transmission/releases/download/latest/flood-for-transmission.tar.gz"
-#   - "mkdir -p /config/flood"
-#   - "tar xf /tmp/flood-for-transmission.tar.gz --strip-components 1 -C /config/flood"
-#   volumeMounts:
-#   - name: config
-#     mountPath: /config
-
 image:
   repository: ghcr.io/k8s-at-home/transmission
   pullPolicy: IfNotPresent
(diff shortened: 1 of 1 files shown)


jr0dd:
> I am going to play with this one day and see if I can get it installed using the initContainer and actaully test. While startupProbe works it's not how it should be used in k8s.

yeah it's definitely hacky. i don't like it, but could not figure out a way to get an initcontainer to get access to the internet.

jr0dd:
A better solution would probably be just adding the tar file to the container image and then using the initcontainer to untar and move to /config

onedr0p:
initContainers _should_ have network access. I am curious what the error logs on your end was?

jr0dd:
> initContainers _should_ have network access. I am curious what the error logs on your end was?

The only output i get is curl unable to resolve github.com before the initcontainer crashes. I tried adding `--dns-servers 1.1.1.1` to the curl command which didn't resolve. I tried multiple images instead of the curl image and same issue.

bjw-s:
I personally feel that startupProbes should not be "abused" for such behavior. A startupProbe is meant to determine if an application has started (hence the name). As @onedr0p states, initContainers _should_ normally have normal networking functionality.

Maybe we can add a test case for this to the `ci` folder so that we can verify if the initContainer works as intended?

onedr0p:
> Maybe we can add a test case for this to the ci folder so that we can verify if the initContainer works as intended?

There is one in the ci values, unless I did something wrong it's been working.

bjw-s:
> There is one in the ci values, unless I did something wrong it's been working.

I must admit I didn't check before mentioning it 😅

jr0dd:
> I personally feel that startupProbes should not be "abused" for such behavior. A startupProbe is meant to determine if an application has started (hence the name). As @onedr0p states, initContainers _should_ normally have normal networking functionality.
> 
> Maybe we can add a test case for this to the `ci` folder so that we can verify if the initContainer works as intended?

Yeah I probably jumped the gun sending a pr for such a hacky thing lol. I'm a little baffled by any initcontainer image I have tried to get outside access. The kubernetes documentation clearly states that initcontainers can be used to pull a git repo. I stuck the curl command in an until loop and let it ran for 15min. Did not resolve. The issue could very well be somewhere obscure on my end. I'll close this pr.

PR #838, opened 2021-04-30
[stable/bitwardenrs] fix ldap labels and names

monotek:
Signed-off-by: André Bauer <monotek23@gmail.com>

**Description of the change**
Changes labels and names of ldap sync resources

**Benefits**
This fixes an issue when the chart, having ldapSync enabled, is installed, uninstalled and installed again, where the ldap secret was not deleted, because label mismatch.

It also fixes an issue when ldapSync is enabled and the chart is installed several times using different release names, where the ldap secret and the deployment got the same name, because the fullname var was not used.

**Possible drawbacks**

**Applicable issues**

- fixes #

**Additional information**

**Checklist** 
- [ ] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [ ] Title of the PR starts with chart name (e.g. `[home-assistant]`)
- [ ] Variables are documented in the README.md (this can be done with using our helm-docs wrapper `./hack/gen-helm-docs.sh stable <chart>`)

Issue #835, opened 2021-04-29
[Bitwarden_rs] Upstream name change


rwaltr:
# Details

**Helm chart name and version:**

Bitwarden_RS upstream is changing their name to [Vaultwarden](https://github.com/dani-garcia/vaultwarden/commit/34ea10475d316ccb2ca4cd2cac67b61c4cdfb62a)

**Container name and tag:**

bitwarden_rs and Vaultwarden

**Relevant Helm values:**

Image and Tags

**What did you expect to happen:**

We need to change our tags and chart name once Vaultwarden's changes reach the registrars

rwaltr:
I'm workiong on a PR

CuBiC3D:
Would be a good chance to migrate to common. @DirtyCajunRice wanted to convert it :D any progress?

rwaltr:
Only basic templates atm. still need to do testing
https://github.com/rwaltr/charts/tree/vaultwarden
It appears that at least these versions of vaultwarden have all their items config thru ENV

PR #841, opened 2021-04-30
[Vaultwarden] New Chart

rwaltr:
**Description of the change**

Adds a new chart to address #835 

Needs to be reviewed by #838 as well to check for Ldap function

**Benefits**

Upstream changed names from Bitwarden_rs to Vaultwarden, used the opportunity to make a commons chart. 

**Possible drawbacks**

Special features from the original chart may be missing. The new chart aims to use the ENV environment for config

**Applicable issues**

#835 

**Additional information**

Bitwarden_RS needs to be deprecated after vaultwarden is deployed and active 

**Checklist** 


charts/stable/vaultwarden/ci/ct-values.yaml
```diff
@@ -0,0 +1,3 @@
+persistence:
+  config:
+    enabled: true
```
Ornias1993, line 3: shouldn't this be an empty-dir?

charts/stable/vaultwarden/values.yaml
```diff
@@ -0,0 +1,34 @@
...
+  IP_HEADER: "X-Real-IP"
+  SIGNUPS_ALLOWED: "true"
+  DOMAIN: "https://bw.domain.tld:8443"
+
+service:
+  port:
+    port: 80
+
+ingress:
+  enabled: false
```
Ornias1993, line 30: It's lacking all the hard-ish ingress config that vaultwarden is especially known for...

charts/stable/vaultwarden/values.yaml
```diff
@@ -0,0 +1,34 @@
...
+env:
+  # TZ:
+  DATA_FOLDER: "config"
+  IP_HEADER: "X-Real-IP"
+  SIGNUPS_ALLOWED: "true"
+  DOMAIN: "https://bw.domain.tld:8443"
+
+service:
+  port:
+    port: 80
```
Ornias1993, line 27: missing the secondary service for websocket connections

charts/stable/vaultwarden/Chart.yaml
```diff
@@ -0,0 +1,20 @@
...
+icon: "https://sesoodan.sirv.com/bitwarden_rs.png"
+sources:
+- https://github.com/dani-garcia/vaultwarden
+maintainers:
+- name: Rwaltr
+  email: rwalt@pm.me
+dependencies:
+- name: common
+  repository: https://library-charts.k8s-at-home.com
+  version: 2.4.0
```
Ornias1993, line 20: Missing optional DB dependencies

Ornias1993 (changes requested):
This has to be compared and have atleast basic feature parity with the current bitwarden_rs charts imho.

While I think some env-vars are quite hard to generate/automate correctly and as such don't really have to be generated/automated at this stage, I think at least basic things like websocket-compatible ingress/services and optional DB-dependencies should be included...

I suggest also looking at:
- https://github.com/gissilabs/charts/tree/master/bitwardenrs
- https://github.com/truecharts/apps/tree/master/stable/bitwarden

charts/stable/vaultwarden/values.yaml
```diff
@@ -0,0 +1,34 @@
...
+  type: Recreate
+
+# See more environment variables in the Vaultwarden documentation
+# https://github.com/dani-garcia/vaultwarden/blob/main/.env.template
+env:
+  # TZ:
+  DATA_FOLDER: "config"
+  IP_HEADER: "X-Real-IP"
+  SIGNUPS_ALLOWED: "true"
+  DOMAIN: "https://bw.domain.tld:8443"
```
onedr0p, line 23: I would comment these out and put them in `vaultwarden/ci/ct-values.yaml` if they required by the CI tests to pass.

onedr0p (changes requested):

charts/stable/vaultwarden/values.yaml
rwaltr: There is a bug with the current chart version that does not allow for any port other than service.port.port to be part of any ingresss

charts/stable/vaultwarden/ci/ct-values.yaml
rwaltr: Why should this be empty dir? That would guarantee config loss. there is more than just templated config here. its also the data dir for Vaultwarden

charts/stable/vaultwarden/values.yaml
rwaltr: There is a bug with the current chart version that does not allow for any port other than service.port.port to be part of any ingresss

charts/stable/vaultwarden/Chart.yaml
rwaltr: will investigate

charts/stable/vaultwarden/values.yaml
rwaltr: `DATA_FOLDER` puts all of vaultwarden's config and data in /config which is conventional for k8s-at-home vs /data

Others are not required

charts/stable/vaultwarden/values.yaml
rwaltr: Added other service

charts/stable/vaultwarden/values.yaml
rwaltr: Workaround found, will apply until new Commons chart deployed

rwaltr:
Added workaround for Web socket ingress until https://github.com/k8s-at-home/library-charts/issues/31

rwaltr:
TIL, Github does not like master rebases

charts/stable/vaultwarden/ci/ct-values.yaml
Ornias1993: This is the testing values.yaml ;-)
It always looses its config :P

charts/stable/vaultwarden/values.yaml
Ornias1993: Yeah I mostly do the same, because it also allows people to have ports as different types (clusterIP vs NodePort)... Not really relevant here, but kind of a "best practice"

charts/stable/vaultwarden/README_CHANGELOG.md.gotmpl
```diff
@@ -0,0 +1,27 @@
...
+All notable changes to this application Helm chart will be documented in this file but does not include changes from our common library. To read those click [here](https://github.com/k8s-at-home/library-charts/tree/main/charts/stable/common#changelog).
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+### [1.0.0]
+
+#### Added
+
+- Initial version
+
```
Ornias1993, line 17:
```suggestion
#### Added

- Initial version, Succeding Bitwarden_RS.

```

charts/stable/vaultwarden/values.yaml
```diff
@@ -0,0 +1,61 @@
...
+  type: Recreate
+
+# See more environment variables in the Vaultwarden documentation
+# https://github.com/dani-garcia/vaultwarden/blob/main/.env.template
+
+mariadb:
+  enabled: false
+
+postgresql:
+  enabled: false
```
Ornias1993, line 23: These are almost always put at the bottom of the Values.yaml file.

Also:
Would be nice to add a comment refering to the BitNami Docs and/or some (commented out) defaults

charts/stable/vaultwarden/values.yaml
```diff
@@ -0,0 +1,61 @@
...
+  additionalIngresses:
+    - enabled: false
+      nameSuffix: "websocket"
+      servicePort: 3012
+      hosts:
+        - host: bw.domain.tdl
+          paths:
+            - path: /notifications/hub
+              pathType: Prefix
+      tls: {}
```
Ornias1993, line 57: You can use a hosttpl to refer to the main ingress hostname for the host, thats a simpler default for most people I think.

Also:
the `notifications/hub/negotiate` should ALSO go to port 80:
`Route everything else, including /notifications/hub/negotiate, to the standard Rocket server, by default at port 80`.

charts/stable/vaultwarden/ci/ct-values.yaml
rwaltr: Thank you, been staring at a lot of text today

rwaltr:
I have reached my limit on this for today, will continue later this week, and perhaps https://github.com/k8s-at-home/library-charts/issues/31 will be ready by then

charts/stable/vaultwarden/values.yaml
```diff
@@ -47,7 +47,7 @@ ingress:
         - host: bw.domain.tdl
           paths:
             - path: /notifications/hub
-              pathType: Prefix
+              pathType: Exact
```
Ornias1993: I don't think this would work, because only one suffix of /notifications/hub needs to go to port 80... This would send all suffixes to 80 instead only the root of /notifications/hub

onedr0p:
2.5.0 common is available 🎉

rwaltr:
Huzzah!

charts/stable/vaultwarden/values.yaml
```diff
@@ -0,0 +1,58 @@
...
+# IMPORTANT NOTE
+#
+# This chart inherits from our common library chart. You can check the default values/options here:
+# https://github.com/k8s-at-home/library-charts/tree/main/charts/stable/common/values.yaml
+#
+
+image:
+  repository: vaultwarden/server
+  pullPolicy: IfNotPresent
+  tag: latest
```
onedr0p, line 11:
```suggestion
  tag: 1.21.0
```

onedr0p (changes requested):
Looks like there is a version tag on dockerhub now

rwaltr:
# TODO
```

postgresql:
  enabled: false
  # postgresqlUsername: ""
  # postgresqlPassword: ""
  # postgresqlDatabase: ""
  # persistence:
  #   enabled: true
  #   storageClass:
  #   accessModes:
  #     - ReadWriteOnce
  ```
  
  bump MariaDB version to 10.5.9
  
  add default https://github.com/bitnami/charts/blob/77bda9977595971ca16ad5ff4c125714f3d79ebf/bitnami/mariadb/values.yaml#L66-L100
  
 set persistence  https://github.com/bitnami/charts/blob/77bda9977595971ca16ad5ff4c125714f3d79ebf/bitnami/mariadb/values.yaml#L315-L318
 
 comment out ingress

PR #879, opened 2021-05-03
[thelounge] Update icon url

nicholaswilde:
**Description of the change**

Updates icon url.

**Benefits**

Updates the icon to one without text.

**Possible drawbacks**

The background of the new icon is tray whereas the background of the current one is black. The current icon may nicer to look at.

**Applicable issues**

- fixes #

**Additional information**

**Checklist** 

### DIFF
--- a/charts/stable/thelounge/Chart.yaml
+++ b/charts/stable/thelounge/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: "4.2.0"
 description: The Lounge, modern web IRC client designed for self-hosting
 name: thelounge
-version: 1.1.1
+version: 1.1.2
 kubeVersion: ">=1.16.0-0"
 keywords:
 - thelounge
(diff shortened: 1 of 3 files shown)


charts/stable/thelounge/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: "4.2.0"
```
carpenike, line 2: Does this need to be quoted? Thought I saw we have it unquoted in other charts.

nicholaswilde: I am not sure about this one. Perhaps @onedr0p or @bjw-s know better?

nicholaswilde: You know, I think I had issues when using two digits for `appVersion`. Helm complained that it didn't want a float and so I had to put it in quotes. I don't think that this applies here.

PR #887, opened 2021-05-05
Do not create SysLog when unifiedService is disabled

### LINKED ISSUE #558: [unifi] The unified service also creates syslog and speedtest services
# Details

unifi 1.5.0

**Container name and tag:**

jacobalberty/unifi:arm32v7

``` yaml
unifiedService:
  enabled: true
```
```sh
helm install -f values.yml -n unifi --create-namespace unifi-controller k8s-at-home/unifi
```

result:
When using the unified service, the syslog and speedtest services are created alongside the unified service.
The port for syslog is available in the unified service but the speedtest port is not.

```sh
NAME                                 TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)                                         AGE
service/unifi-controller-speedtest   ClusterIP   10.43.67.177    <none>        6789/TCP                                        25m
service/unifi-controller-syslog      NodePort    10.43.118.162   <none>        5514:32115/UDP                                  25m
service/unifi-controller             ClusterIP   10.43.194.125   10.0.0.101    8080/TCP,10001/UDP,3478/UDP,5514/UDP,8443/TCP   25m
```

The syslog and speedtest services should not be created.  

Its most likely the if statement on line 1 in syslog-svc.yml and speedtest-svc.yml:
```yaml
{{ if not .Values.syslogService.enabled }}
```

```yaml
{{ if not .Values.speedtestService.enabled }}
```

 should be:
```yaml
{{ if not .Values.unifiedService.enabled }}
```

**Additional Information:**

None

beshkenadze:
**Description of the change**

Do not create SysLog when unifiedService is disabled. 

**Benefits**

SysLog service would not be created when disabled. 

**Applicable issues**

- fixes #558

**Additional information**

**Checklist** 


bjw-s:
Thanks for taking the time to submit this PR. We would appreciate it if you would next time fill out our PR template. That would help us to better assess the scope and impact of a PR.

As for the change, apart from the chart metadata that hasn't been updated (chart version, changelog, etc), I am curious why would want to modify the syslog service behavior to be different from all the other services? The logic there is also: if the unifiedService is *not* enabled, create separate services.

beshkenadze:
@bjw-s oh, I forgot to update the chart metadata, sorry.
So, about services: it's not obvious behavior and also broke installing from Helm chart with default values.

charts/stable/unifi/templates/syslog-svc.yaml
```diff
@@ -1,4 +1,4 @@
-{{ if not .Values.unifiedService.enabled }}
+{{ if .Values.syslogService.enabled }}
```
bjw-s, line 1:
```suggestion
{{ if (not .Values.unifiedService.enabled) and .Values.syslogService.enabled }}
```

bjw-s: I think this makes the behavior more explicit, and won't allow two services for the same thing. Maybe it would be best to apply this style of logic to all the non-unified services.

bjw-s (changes requested):

charts/stable/unifi/templates/syslog-svc.yaml
beshkenadze: Syslog is used for remote logging and not necessary.
I also see unifiedService is dependent on Syslog, speedtestService, etc., in unified-svc.yaml.
In my opinion, the unified services should separate necessary services and optionally services.

bjw-s:
Sorry, my suggestion broke some stuff (missing some parentheses) :( I'll push a fix for that in a bit.

bjw-s (dismissed):

bjw-s:
@all-contributors add @beshkenadze for code

PR #912, opened 2021-05-15
[multus] add Multus

angelnu:
**Description of the change**

Multus allows having multiple NICs in a POD. This provides a helm chart
that installs multus and the extra CNI plugins missing in k3s.

**Benefits**

Can have PODs access the home network without network=host which has
security and port conflict issues. Accessing the home network allows
DDNS (Google home) and other protocols to work that do not propagate
through routers.

**Possible drawbacks**

NA

**Applicable issues**

- fixes NA

**Additional information**

**Checklist** 

### DIFF
--- a/.github/ct-install.yaml
+++ b/.github/ct-install.yaml
@@ -7,6 +7,7 @@ chart-dirs:
 excluded-charts:
 - charts/stable/alertmanager-bot
 - charts/stable/dnsmadeeasy-webhook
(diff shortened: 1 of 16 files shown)


angelnu:
Testcase disabled - the ks3 deployed by the workflow does not have /var/lib/rancher/k3s/data/current/bin and therefore fails to install.

onedr0p (approved):

PR #923, opened 2021-05-24
[maddy] add new chart

ChipWolf:
**Description of the change**

Added a chart for [maddy](https://github.com/foxcpp/maddy)

**Benefits**

Support the only chart for maddy

**Possible drawbacks**

Additional maintenance overhead

**Additional information**

**Checklist** 


ChipWolf:
Right, so the container for maddy isn't great.
It explains in the documentation the container will fail on first startup, so I'll need to preempt the content of the volume with this chart

rwaltr:
> Right, so the container for maddy isn't great.
> It explains in the documentation the container will fail on first startup, so I'll need to preempt the content of the volume with this chart

perhaps use an initcontainer for this?

ChipWolf:
I'm no longer using this software, so have no need to continue working on the chart, if someone wants to take over feel free.

PR #931, opened 2021-06-03
[pod-gateway] Rename chart since it includes gateway now

angelnu:
**Description of the change**

Renamed `pod-gateway-setters` to `pod-gateway`.

The original idea was to split the webhook into a dedicated char separated from the gateway. This did not work out
since the webhook inserted containers and the gateway depend on each other. This is because they need to
create a VXLAN overlay to route the traffic on through all the K8S routers.

**Benefits**

Single chart that contains all required code.

**Possible drawbacks**

NA

**Applicable issues**

- fixes NA

**Additional information**

I did not update the chart version since I renamed the chart.

**Checklist** 


charts/stable/pod-gateway/README_CHANGELOG.md.gotmpl
```diff
@@ -13,7 +13,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 #### Added
 
-- Initial version
+- renamed chart
```
bjw-s, line 16: This seems "weird" for a new chart... Maybe mention here as well that this is the successor to `pod-gateway-setter`?

angelnu: done, thanks for the review.

bjw-s (approved):

PR #936, opened 2021-06-04
[pod-gateway] Support selective configmap generation

angelnu:
**Description of the change**

Generate the configmap out of yaml so that only the required override values are required. This is also needed
to ensure that when new options get added we do not affect existing users.

**Benefits**

More robust adding new features

**Possible drawbacks**

A disruptive change

**Applicable issues**

- fixes NA

**Additional information**

**Checklist** 


bjw-s (commented):
Since it's a breaking release anyway, how would you feel about instead of `.Values.configmap.settings` just making it `.Values.settings`? 

Is there an option to specify an existing configMap? Because otherwise you could also always create it. This would then also remove the need for the `configmap` key in values.yaml.

angelnu:
I missed your feedback this morning @bjw-s - added. Hope it works for your as well :-)

bjw-s (approved):

PR #937, opened 2021-06-05
[pod-gateway] fix: run gateway init as privileged

toboshii:
**Description of the change**

This sets the gateway init container to `privileged: true` to support https://github.com/k8s-at-home/pod-gateway/pull/5. This is required for CNIs that disable IP forwarding irrespective of the host node's settings (Calico, possibly others.)

**Benefits**

Allows the gateway to work with Calico

**Possible drawbacks**

Not great security wise, but the other option is forcing users to enable container forwarding globally in the CNI settings.

**Applicable issues**

**Additional information**

**Checklist** 


toboshii:
Merged into #936 downstream

PR #1011, opened 2021-06-14
[leaf2mqtt] Migrate to common v3

mitsumaui:
**Description of the change**

Update leaf2mqtt to use new common chart 3.1.0

**Benefits**

uses new common chart

**Possible drawbacks**

**Applicable issues**

**Additional information**


charts/stable/leaf2mqtt/Chart.yaml
```diff
@@ -21,4 +21,4 @@ maintainers:
 dependencies:
 - name: common
   repository: https://library-charts.k8s-at-home.com
-  version: 2.5.0
+  version: 3.1.0
```
bjw-s, line 24:
```suggestion
  version: 3.1.1
```

charts/stable/leaf2mqtt/values.yaml
```diff
@@ -32,9 +32,6 @@ env: {}
   # CHARGING_UPDATE_INTERVAL_MINUTES:
   # LOG_LEVEL:
 
-service:
```
bjw-s, line 35: Is there a UI/similar service that needs to be exposed? If so, a port number will need to be specified.

bjw-s (changes requested):
Thanks for creating this PR 👍! Before we approve the PR, could you please update the chart's documentation (changelog and helm-doc strings, and regenerate the README) as well?

charts/stable/leaf2mqtt/Chart.yaml
mitsumaui: Ah so we prefer to follow the common chart version where possible? That's fine

charts/stable/leaf2mqtt/values.yaml
mitsumaui: No Service / listeners, only outbound communication.

mitsumaui:
> Thanks for creating this PR +1! Before we approve the PR, could you please update the chart's documentation (changelog and helm-doc strings, and regenerate the README) as well?

Now - I thought there was somewhere this happened automagically - or was in a task? Is there something I'm missing on how to manage / update the docs as failing to find it in the kah chart docs?

charts/stable/leaf2mqtt/Chart.yaml
mitsumaui: Ah - ignore this going crazy :)

bjw-s (changes requested):

bjw-s:
I've gone ahead and migrated the chart in #1077. Still thanks a lot for the time you put in to this PR!

mitsumaui:
> I've gone ahead and migrated the chart in #1077. Still thanks a lot for the time you put in to this PR!

Thanks for sorting that - apologies I didn't finish this out!

Issue #942, opened 2021-06-08
Migrate charts to common v3


bjw-s:
# Details

The following charts need to be upgraded to the latest version of the common library. Since this is a mostly manual task, we'll track the progress here 

#### To Do:

- [ ] baikal (#1077)
- [ ] blocky (#1077)
- [ ] comcast (#1077)
- [ ] gollum (#1077)
- [ ] jetbrains-projector (#1077)
- [ ] kanboard (#1077)
- [ ] leaf2mqtt (#1011)
- [ ] unpackerr (#1077)
- [ ] uptimerobot-prometheus (#1077)
- [ ] valheim (#1077)
- [ ] vikunja (#1077)
- [ ] wallabag (#1077)
- [ ] whoogle (#1077)
- [ ] wikijs (#1077)
- [ ] xbackbone (#1077)
- [ ] youtubedl-material (#1077)

#### Done:

- [x] ~adguard-home (#952)~
- [x] ~airsonic (#941)~
- [x] ~alertmanager-bot (#954)~
- [x] ~appdaemon (#1070)~
- [x] ~bazarr (#969)~
- [x] ~booksonic-air (#989)~
- [x] ~bookstack (#950)~
- [x] ~calibre (#1004)~
- [x] ~calibre-web (#973)~
- [x] ~deconz (#1017)~
- [x] ~deluge (#981)~
- [x] ~dizquetv (#984)~
- [x] ~dnsmadeeasy-webhook (#986)~
- [x] ~dsmr-reader (#968)~
- [x] ~duplicati (#987)~
- [x] ~emby (#961)~
- [x] ~esphome (#988)~
- [x] ~flaresolverr (#990)~
- [x] ~flood (#996)~
- [x] ~focalboard (#997)~
- [x] ~founderyvtt (#948)~
- [x] ~freshrss (#998)~
- [x] ~gaps (#999)~
- [x] ~gonic (#1000)~
- [x] ~grocy (#1001)~
- [x] ~haste-server (#947)~
- [x] ~healthchecks (#1006)~
- [x] ~heimdall (#1002)~
- [x] ~home-assistant (#964)~
- [x] ~homebridge (#1005)~
- [x] ~homer (#1007)~
- [x] ~hyperion-ng (#1008)~
- [x] ~icantbelieveitsnotvaletudo (#1009)~
- [x] ~intel-gpu-plugin (#966)~
- [x] ~jackett (#992)~
- [x] ~jellyfin (#977)~
[...]

PR #1044, opened 2021-06-21
feat: migrate multiple charts to common v3

onedr0p:
**Description of the change**

Migrate the following charts to common v3
- nzbhydra2
- overseerr
- sabnzbd
- tautulli
- xteve
- traccar
- stash

Related: https://github.com/k8s-at-home/charts/issues/942

### DIFF
--- a/charts/stable/stash/ci/ct-values.yaml
+++ b/charts/stable/stash/ci/ct-values.yaml
@@ -11,8 +11,5 @@ persistence:
     mountPath: /root/.stash
   media:
     enabled: true
-    emptyDir:
-      enabled: true
-    size: 10M
+    type: emptyDir
     mountPath: /media
-    accessMode: ReadWriteOnce
--- a/charts/stable/stash/ci/ct-values.yaml
+++ b/charts/stable/stash/ci/ct-values.yaml
@@ -10,8 +10,10 @@ persistence:
     emptyDir:
       enabled: true
     mountPath: /root/.stash
+    accessMode: ReadWriteOnce
   media:
     enabled: true
     emptyDir:
       enabled: true
     mountPath: /media
+    accessMode: ReadWriteOnce
--- a/charts/stable/stash/ci/ct-values.yaml
+++ b/charts/stable/stash/ci/ct-values.yaml
@@ -9,11 +9,13 @@ persistence:
     enabled: true
     emptyDir:
       enabled: true
+    size: 10M
     mountPath: /root/.stash
     accessMode: ReadWriteOnce
   media:
     enabled: true
     emptyDir:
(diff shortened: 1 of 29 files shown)


charts/stable/stash/ci/ct-values.yaml
```diff
@@ -9,9 +9,13 @@ persistence:
     enabled: true
     emptyDir:
       enabled: true
+    size: 10M
     mountPath: /root/.stash
+    accessMode: ReadWriteOnce
```
bjw-s, line 14: Why not use a `type: emptyDir`? this of course is the next best thing, but then the `emptyDir` key can be removed completely :)

charts/stable/stash/ci/ct-values.yaml
```diff
@@ -9,9 +9,13 @@ persistence:
...
   media:
     enabled: true
     emptyDir:
       enabled: true
```
bjw-s, line 18: Same as previous comment :)

bjw-s (changes requested):
Only 2 comments (which is actually just 1 comment/question) on all these files, that's pretty great 🔥

charts/stable/stash/ci/ct-values.yaml
onedr0p: Not sure I understand, I was following how the paperless PR was handling it.

bjw-s: I noticed... I wanted to comment / ask the question on that as well. In other charts that require some kind of persistence during the `ct install` CI phase, we just give it an emptyDir instead of a real PVC. That way we don't have to rely on a CSI being present. So it would become something like this:

```yaml
persistence:
  config:
    enabled: true
    type: emptyDir
```

bjw-s, line 14:
```suggestion
    type: emptyDir
    mountPath: /root/.stash
```

charts/stable/stash/ci/ct-values.yaml
```diff
@@ -9,9 +9,13 @@ persistence:
...
+    size: 10M
     mountPath: /media
+    accessMode: ReadWriteOnce
```
bjw-s, line 21:
```suggestion
    type: emptyDir
    mountPath: /media
```

bjw-s (changes requested):

charts/stable/stash/ci/ct-values.yaml
bjw-s: I spoke with @angelnu about this, and we can continue to use emptyDir's 🙂

bjw-s (approved):

PR #1061, opened 2021-06-24
[pod-gateway] Fix issue where values.yaml defined 'configFile' wouldn't work in vpn addon

jmmaloney4:
**Description of the change**

Because the default values.yaml file has `configFileSecret` set to `openvpn` by default, it is impossible to use the built in `configFile` with no secret. If you attempt to override `configFileSecret` to `null` the default value takes precedence, and the gateway will not start because of the lack of `openvpn` secret. 

**Benefits**

**Possible drawbacks**

**Applicable issues**

- fixes #

**Additional information**

**Checklist** 

### DIFF
--- a/charts/stable/pod-gateway/values.yaml
+++ b/charts/stable/pod-gateway/values.yaml
@@ -76,7 +76,7 @@ addons:
     openvpn:
     wireguard:
     env:
-    configFileSecret: openvpn
+    # configFileSecret: openvpn
     networkPolicy:
       enabled: true
       egress:
--- a/charts/stable/pod-gateway/values.yaml
+++ b/charts/stable/pod-gateway/values.yaml
@@ -73,9 +73,9 @@ addons:
     # for extra safeness in case the VPN does connect
     enabled: false
     type: openvpn
-    openvpn:
-    wireguard:
-    env:
+    # openvpn:
+    # wireguard:
+    # env:
     # configFileSecret: openvpn
     networkPolicy:
       enabled: true
--- a/charts/stable/pod-gateway/Chart.yaml
+++ b/charts/stable/pod-gateway/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 1.2.6
 description: Admision controller to change the default gateway and DNS server of PODs
 name: pod-gateway
-version: 3.2.1
+version: 3.2.2
 kubeVersion: ">=1.16.0-0"
 keywords:
 - pod-gateway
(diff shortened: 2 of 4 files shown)


bjw-s:
In my local testing, when you specify the following values you should be able to run without the configFile:

```yaml
addons:
  vpn:
    enabled: true
    type: openvpn
    openvpn:
    configFileSecret: 
```

Although I do think it would be a bit cleaner if some of these keys are not set here but are left to the user's preference. Unfortunately I ran into a small bug in the common library when trying this, so that would need to get fixed before we can remove the `openvpn` and `wireguard` keys.

angelnu:
Thanks for the PR!

The values I picked as the defaults were heavily influenced by me using openvpn - so we can remove any non-default values that cannot be overwritten by the user.

Is there anything to remove beyond `configFileSecret` and potentially `type` ? Is there any issue with the null values @bjw-s or are they merged with the common defaults?

bjw-s:
> Is there anything to remove beyond `configFileSecret` and potentially `type` ? Is there any issue with the null values @bjw-s or are they merged with the common defaults?

By specifying them as null-values, the default values seem to get overwritten with `null` (I still can't really figure out why :X), potentially causing the chart to render incorrectly. That's why I proposed to remove the `wireguard` and `openvpn` keys.

charts/stable/pod-gateway/values.yaml
```diff
@@ -76,7 +76,7 @@ addons:
     openvpn:
```
angelnu, line 76:
```suggestion
    # openvpn:
```

charts/stable/pod-gateway/values.yaml
```diff
@@ -76,7 +76,7 @@ addons:
...
     wireguard:
```
angelnu, line 77:
```suggestion
    # wireguard:
```

charts/stable/pod-gateway/values.yaml
```diff
@@ -76,7 +76,7 @@ addons:
...
     env:
```
angelnu, line 78:
```suggestion
    # env:
```

angelnu (changes requested):

angelnu:
@jmmaloney4 - please merge the suggestions when you have the chance so we can merge your PR. Thanks again for submitting 👍

angelnu:
@jmmaloney4 - would you be able to add the docs? If you need any help please let me know - I can, if necessary, add the doc to your PR so we can merge this fix.

jmmaloney4:
@angelnu is this change sufficient?

angelnu:
It looks good @jmmaloney4 - only step left is to re-generate the README.md.

Can you please run `task chart:helm-docs CHART_TYPE=stable CHART=pod-gateway` in your PR? Then we are ready to merge :-)

jmmaloney4:
@angelnu should be good to go.

angelnu (approved):

angelnu:
Thanks @jmmaloney4 - now we need an owner to merge.

bjw-s (approved):
Approved! Thanks a lot for the contribution!!

bjw-s:
@all-contributors please add @jmmaloney4 for code

PR #1091, opened 2021-07-18
[frigate] chart based on common

onedr0p:
**Description of the change**

Adding Frigate back into our Charts Library, starting at version 5.0.0 due to last Frigate chart was 4.x

**Benefits**

While there is an [official Chart](https://github.com/blakeblackshear/blakeshome-charts) for Frigate, I feel we can do a better job supporting it here using the common library. We will be responsive to changes as the app changes overtime.

**Additional information**

**Checklist** 


charts/stable/frigate/templates/configmap.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "common.names.fullname" . }}-config
+  labels:
+    {{- include "common.labels" . | nindent 4 }}
+data:
+  config.yml: |
+    {{- .Values.config | nindent 4 }}
```
bjw-s, line 9: Could be nice to allow templating here
```suggestion
    {{- .Values.config | tpl | nindent 4 }}
```

onedr0p: Good point! Added.

charts/stable/frigate/values.yaml
```diff
@@ -0,0 +1,106 @@
...
+
+ingress:
+  # -- Enable and configure ingress settings for the chart under this key.
+  # @default -- See values.yaml
+  main:
+    enabled: false
+
+securityContext:
+  # -- (bool) Privileged securityContext may be required if USB devices are accessed directly through the host machine
+  privileged: true
```
bjw-s, line 46:
```suggestion
  privileged:  # true
```
Missed this one earlier, in other charts we have the user explicitly set this:
https://github.com/k8s-at-home/charts/blob/09ab3726253bffb2b9bd7f6afc8bb129d0a4d285/charts/stable/zigbee2mqtt/values.yaml#L38

onedr0p: I just set it to false by default, that ok?

bjw-s (approved):

PR #1102, opened 2021-07-29
[blocky] Move replicas under controller definition in values.yaml

brandon099:
Signed-off-by: Brandon Clifford <brandon@clifford.sh>

**Description of the change**

This moves the `replicas` underneath `controller` as the move to the current common chart expects it here.

**Benefits**

Update's self-documentation mentioned in the values.yaml to indicate where replicas should be defined.

**Checklist** 


bjw-s (approved):

bjw-s:
Good catch! Thanks for taking the time to submit a PR!

PR #1111, opened 2021-08-03
[frigate] Adding more detail to values.yaml

### LINKED ISSUE #1109: Frigate chart version 5.0.0's value.yaml is confusing around persistence
The frigate chart's readme say to look into the value.yaml file for information about persistence, but the [values.yaml ](https://github.com/k8s-at-home/charts/blob/2df5b026b8c08f3aa1818119eefa500c0b401bd8/charts/stable/frigate/values.yaml#L49)has no real information about persistence inside.

```
# @default -- See values.yaml
persistence:
  data:
    enabled: false
    mountPath: /data
  media:
    enabled: false
    mountPath: /media
  # -- Configure a hostPathMount to mount a USB device in the container.
  # @default -- See values.yaml
  usb:
    enabled: false
    type: hostPath
    hostPath: /dev/bus/usb
  # -- Configure a temporary cache. See [docs](https://blakeblackshear.github.io/frigate/) for more details..
  # @default -- See values.yaml
  cache:
    enabled: true
    type: emptyDir
    medium: Memory
    sizeLimit: 1Gi
    mountPath: /dev/shm
```
Specifically, I was looking to understand what is data vs media. Would I need both? Do I just need one? What gets written to media vs data? This may be something to pass off to the container code like the temporary cache.

The top comment also seems to refer to itself.

hyperbolic2346:
I'm not married to these changes, but wanted to show the kind of changes I was wanting to resolve #1109

- fixes #1109 

**Checklist** 

### DIFF
--- a/charts/stable/frigate/Chart.yaml
+++ b/charts/stable/frigate/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: v0.8.0
 description: NVR With Realtime Object Detection for IP Cameras
 name: frigate
-version: 5.0.0
+version: 5.0.1
 kubeVersion: ">=1.16.0-0"
 keywords:
 - frigate
--- a/charts/stable/frigate/values.yaml
+++ b/charts/stable/frigate/values.yaml
@@ -23,7 +23,7 @@ env:
   FRIGATE_RTSP_PASSWORD: ""
 
 # -- Configures service settings for the chart.
-# @default -- See values.yaml
+# @default -- library chart See values.yaml
 service:
   main:
     ports:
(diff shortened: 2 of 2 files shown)


bjw-s:
Thanks for taking the time to create the PR. The annotations that you have changed are part of the rendering of our chart README files ([by helm-docs](https://github.com/norwoodj/helm-docs#default-valuescolumn)).

So whilst I can agree that they may look a bit weird when seeing them within the `values.yaml` file they do serve a purpose.

We also try (but sometimes fail 😉) to keep the wording consistent throughout our charts, so we won't be changing this for now.

hyperbolic2346:
How about the comments regarding the persistence?

What brought me to talk about the values.yaml was that I looked at the readme which said look at values.yaml and in there it said very little. Would these changes be better served inside the readme? Are the helm docs rendered somewhere else that I didn't find?

PR #1112, opened 2021-08-06
[youtubedl-material] Fix incorrect mount paths

ryanwholey:
**Description of the change**

The change updates the pv mount paths to what I believe to be the correct paths. The app reads subscriptions from `/app/subscriptions`, videos from `/app/videos`, audio from `/app/audio` and config from `/app/appdata`. The way that the mount paths are written in the values.yaml, the volumes for subscriptions and video mount at the root directory instead of under /app.

I noticed there was a discrepancy with the mount paths when my videos were gone due to a pod eviction. The eviction was due to an exceeded ephemeral storage limit which makes sense because I was unknowingly storing videos within the container. 

As a workaround, I'm currently passing the mountPaths as values to my installation and I'm able to persist videos and audio as expected. 

**Benefits**

The installation will store videos and audio into the volumes that one would expect when enabling them.

**Possible drawbacks**

API change, although if the values don't work as expected then it might be a welcomed change

**Applicable issues**

None, but I can open one if necessary

**Additional information**

There is a volume for `users` at the mount path `/users`, but it doesn't seem like it should exist. Users are stored at `/app/appdata/users.json` which would be under the `config` volume.

**Checklist**

### DIFF
--- a/charts/stable/youtubedl-material/Chart.yaml
+++ b/charts/stable/youtubedl-material/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: "4.2"
 description: Self-hosted YouTube downloader built on Material Design
 name: youtubedl-material
-version: 3.1.0
+version: 3.1.1
 kubeVersion: ">=1.16.0-0"
 keywords:
 - youtubedl-material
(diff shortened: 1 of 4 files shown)


bjw-s (dismissed):

ryanwholey:
@bjw-s sorry about that! Got the chart version, missed the docs/changelog. Hopefully good to go now 🙏

bjw-s:
No worries, thanks for the quick fix :) Merging! 🍻

bjw-s (approved):

PR #1117, opened 2021-08-10
[recipes] Nginx sidecar mime.types fix

zhorvath83:
**Description of the change**

- Configuring Nginx sidecar MIME types for serving `Recipes` static content.

- With the current Nginx sidecar config static mount persistence is mandatory so it is the new default value. 

**Benefits**

Recipes static content will be served correctly with proper content type.

**Possible drawbacks**

Nothing.

**Applicable issues**

- With the first fix previous Nginx configuration Recipes won't load properly, because the static content (eg. CSS) is served with content type: 'text/plain'. 

- Without the second fix Nginx throws http 500 status codes when serving `Recipes` static content (eg. CSS).

It causes the undermentioned error:
> Refused to apply style from '<URL>' because its MIME type ('text/plain') is not a supported stylesheet MIME type, and strict MIME checking is enabled.

**Checklist** 
- [X] Variables are documented in the README.md


zhorvath83:
I made additional volume fix and I made up the changelog.

bjw-s:
Thanks for updating the PR! I'm going to merge this into the branch that contains our newest `common` version, so we only have to bump the version once for both changes :)

PR #1125, opened 2021-08-18
[octoprint] new images, configmap, init container & much more [WIP]

### LINKED ISSUE #1118: [Octoprint] Error: Invalid value for '--port': tcp://10.43.245.198:80 is not a valid integer
# Details

**Helm chart name and version:** Octoprint 6.0.0

**Container name and tag:** "octoprint/octoprint:1.6.1"

**What steps did you take and what happened:** Deployed with nearly default values.yaml

**Relevant Helm values:**
```yaml
env:
  TZ: UTC
  ENABLE_MJPG_STREAMER: "true"
  MJPG_STREAMER_INPUT:  # "-y -n -r 640x480"
  CAMERA_DEV:  # /dev/video0

ingress:
  main:
    enabled: true
    annotations:
      cert-manager.io/cluster-issuer: vault-issuer
    hosts:
    - host: octoprint.dc
      paths:
      - path: /
        pathType: Prefix
    tls:
    - secretName: octoprint-tls
      hosts:
      - octoprint.dc

securityContext:
  privileged: true

persistence:
  data:
    enabled: true
    mountPath: /octoprint
    storageClass: local-path
    accessMode: ReadWriteOnce
    size: 1Gi
  printer:
    enabled: true
    type: hostPath
    hostPath: /dev/ttyACM0
```

**What did you expect to happen:** Start Octoprint

It's clear that he just expects the port number, but i couldn't figure out where it is set & why he is using the service / cluster] IP.

**Additional Information:**
Full Error output:
```
Error: Invalid value for '--port': tcp://10.43.245.198:80 is not a valid integer
Usage: octoprint [OPTIONS] COMMAND [ARGS]...
Try 'octoprint --help' for help.
```
Also it seems to be missing live/readiness probes.

Nold360:
**Description of the change**

The current octoprint chart uses a image the contains multiple services e.g. `haproxy`, `octoprint` and `mjpg-streamer`. This change brakes up the monolithic image into multiple containers inside a pod, using a `minimalistic` upstream image.

Also I added a few configuration options directly to the `values.yaml` making it easy to use. Besides `mjpg-streamer` it's also possible to use `ustreamer` by simply enabling it in the `values.yaml`

Furthermore I had to create a initContainer which installes octoprint, since octoprint gets installed into `/octoprint` during the creation of the image & will be empty if a persistent volume is being used. While at it I also implemented an automatic installation & upgrade of octoprint plugins during init phase.

**Benefits**
 - Better practice by separating services in multiple containers
 - ustreamer support
 - less overhead by using ingress instead of haproxy
 - more & simple configuration options for [webcam streamers](https://github.com/Nold360/k8s-at-home-charts/blob/d88c7eb3828527aa04dac5732957ddc2d2453fa2/charts/stable/octoprint/values.yaml#L59) / octoprint /...
 - [easy automatic install of octoprint plugins](https://github.com/Nold360/k8s-at-home-charts/blob/d88c7eb3828527aa04dac5732957ddc2d2453fa2/charts/stable/octoprint/values.yaml#L45) via init-container & array in `values.yaml`

**Possible drawbacks**
 - Ingress is needed to use a the webcam streamer [well i guess you could also use a NodePort...]
 - octoprint installation of first start might require internet access, due to pip. [But all dependencies should already be installed anyways & i'm using the original .tar.gz which is still inside the container image]
 - Probably breaks current deployments 

**Applicable issues**

- fixes #1118

**Additional information**

I marked this PR as "WIP" since 
  1. I want your feedback 
  2. I need a better way to implement the second path on the ingress. Since the name of the streamer-service is dynamic ( Release.Name-Chart.Name-myService ). I tried adding it to the paths-array like i do with the `additionalContainers`. But since it's an array (in an array) i wasn't able to make it work. [See this](https://github.com/Nold360/k8s-at-home-charts/blob/d88c7eb3828527aa04dac5732957ddc2d2453fa2/charts/stable/octoprint/values.yaml#L110)

I guess the easiest way would be to create a Service-template with a fixed name & keep the ingress path in `values.yaml`

**Checklist** 

### DIFF
--- a/charts/stable/octoprint/templates/common.yaml
+++ b/charts/stable/octoprint/templates/common.yaml
@@ -1,1 +1,34 @@
+{{/* Make sure all variables are set properly */}}
+{{- include "common.values.setup" . }}
+
+{{/* Append the configMap volume to the volumes */}}
+{{- define "octoprint.init" -}}
+enabled: "true"
+mountPath: "/init.sh"
+subPath: "init.sh"
+type: "custom"
+volumeSpec:
+  configMap:
+    name: {{ include "common.names.fullname" . }}-init
+    defaultMode: 0755
+{{- end -}}
+{{- define "octoprint.initContainer" -}}
+name: init-octoprint
+image: {{ .Values.image.repository }}:{{ .Values.image.tag }}
+imagePullPolicy: {{ .Values.image.pullPolicy }}
+volumeMounts:
+- name: data
+  mountPath: /octoprint
+- name: octoprint-init
+  mountPath: /init.sh
+  subPath: init.sh
+command:
+  - sh
+  - -c
+  - /init.sh
+{{- end -}}
+{{- $_ := set .Values.persistence "octoprint-init" (include "octoprint.init" . | fromYaml) -}}
+{{- $_ := set .Values.initContainers "octoprint-init" (include "octoprint.initContainer" . | fromYaml) -}}
+
+{{/* Render the templates */}}
 {{ include "common.all" . }}
--- a/charts/stable/octoprint/templates/init_config.yaml
+++ b/charts/stable/octoprint/templates/init_config.yaml
@@ -1,0 +1,17 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ template "common.names.fullname" . }}-init
+  labels:
+    {{- include "common.labels" . | nindent 4 }}
+data:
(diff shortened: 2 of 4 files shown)


bjw-s:
Thanks for the work you've been doing on this! I'm still away on vacation for another week or so, but I'll definitely take a look and see where I can help when I get back.

PR #1129, opened 2021-08-21
[smarter-device-manager] use /var/log for termination.messagePath (fixes #1092)

### LINKED ISSUE #1092: [smarter-device-manager] Pods can not start with default configuration
# Details

smarter-device-manager 5.1.0

**Container name and tag:**

registry.gitlab.com/arm-research/smarter/smarter-device-manager:v1.20.7

`helm install smarter-device-manager k8s-at-home/smarter-device-manager`

Pods could not start due to attempting to use `/dev/termination-log` for `terminationMessagePath` while default configuration causes `/dev/` to be mounted as read-only

**Relevant Helm values:**

Defaults only used

Pods start successfully

Was able to get pods to start by setting:

```yaml
persistence:
  devfs:
    enabled: true
    type: hostPath
    hostPath: /dev
    readOnly: false
  sysfs:
    enabled: true
    type: hostPath
    hostPath: /sys
    readOnly: true
  kubeletsockets:
    enabled: true
    type: hostPath
    hostPath: "/var/lib/kubelet/device-plugins"
```

I believe that the better solution is to change the path for `terminationMessagePath`. 

The common helm charts, however, do not seem to support this capability

**Additional Information:**

zamnuts:
Signed-off-by: Andrew Zammit <zammit.andrew@gmail.com>

**Description of the change**

- Override the default container `terminationMessagePath` to allow mounting `/dev` as read-only. The path is now `/var/log/termination-log`. Fixes #1092.
- Use the container's error log to populate the container termination message, since smarter-device-manager does not support a k8s `termination-log`.

**Benefits**

The default `values.yaml` for this chart now work out-of-the-box.

**Possible drawbacks**

This chart's `common` dependency was bumped on 8/12 to v4.0.0, which added support for the `termination` section. The default values added here may conflict with the consumer's expectation of behavior, albeit it never worked to begin with. If set by the consumer, these default values will have no effect.

**Applicable issues**

- fixes #1092 

**Additional information**

**Checklist** 

### DIFF
--- a/charts/stable/smarter-device-manager/Chart.yaml
+++ b/charts/stable/smarter-device-manager/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 1.20.7
 description: Manage hardware resource allocation without a need for privileged containers
 name: smarter-device-manager
-version: 6.0.0
+version: 6.1.0
 kubeVersion: ">=1.16.0-0"
 keywords:
 - kubernetes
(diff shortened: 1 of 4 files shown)


onedr0p (approved):

onedr0p:
Thanks for the fix!

PR #1130, opened 2021-08-21
[adguard-home] Add pullPolicy for configmap init

ElGranLoky:
If adguard is not working no dns, best use IfNotPresent for configmap init.

### DIFF
--- a/charts/stable/adguard-home/Chart.yaml
+++ b/charts/stable/adguard-home/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: v0.106.3
 description: DNS proxy as ad-blocker for local network
 name: adguard-home
-version: 5.0.0
+version: 5.0.1
 kubeVersion: ">=1.16.0-0"
 keywords:
 - adguard-home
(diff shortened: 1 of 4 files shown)


onedr0p:
I'm alright with this change but please bump the chart version and re-run helm-docs. Thanks!

onedr0p (approved):

onedr0p:
@ElGranLoky I just wanted to say your k8s nodes should not be relying on Adguard for DNS. Therefore you should not lose DNS during a update.

ElGranLoky:
Thanks @onedr0p.

I fix it in my deployment. I added a public dns in secundary nameserver in the nodes.

I was also thinking about the rate limit from pull request in docker hub.

onedr0p:
@ElGranLoky adding a secondary dns that's public doesn't matter. When DNS requests are looked up it randomly chooses one and if it fails it doesn't fall back to the secondary DNS. It's wise to only put one DNS provider. See some of my guidelines here. https://onedr0p.github.io/home-cluster/installation/guidelines/

Also, you dockerhub rate limit is another issue but we could try and replace BusyBox with a image not hosted there.

ElGranLoky:
@onedr0p Thanks for the guideline

PR #1135, opened 2021-08-24
[wikijs] Add fsGroup and sort persistent volumes. Add DB_HOST.

fezzik02:
**Description of the change**

This change adds a clearer persistentVolume config, fixes the fsGroup setting to make sure persistent volumes will be writeable, and makes it clear how to configure an external database.

**Benefits**

Better, more consistent deploy process.

**Possible drawbacks**

Possibly redundant disk layout if lsio is symlinking behind our backs

**Applicable issues**

- fixes #

**Additional information**

Not tested but based on the config that I used to get it to work.
**Checklist** 

### DIFF
--- a/charts/stable/wikijs/Chart.yaml
+++ b/charts/stable/wikijs/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: version-2.5.201
 description: Make documentation a joy to write using Wiki.js's beautiful and intuitive interface!
 name: wikijs
-version: 6.0.0
+version: 6.0.1
 kubeVersion: ">=1.16.0-0"
 keywords:
 - wiki
--- a/charts/stable/wikijs/README.md
+++ b/charts/stable/wikijs/README.md
@@ -78,6 +78,7 @@ N/A
 |-----|------|---------|-------------|
 | env | object | See below | environment variables. See [image docs](https://docs.linuxserver.io/images/docker-wikijs#environment-variables-e) for more details. |
 | env.DB_FILEPATH | string | `"/app/wiki/data/db.sqlite"` | Path to the SQLite file |
+| env.DB_HOST | string | `nil` | Host of SQL server |
 | env.DB_NAME | string | `nil` | Database name |
 | env.DB_PASS | string | `nil` | Password to connect to the database |
 | env.DB_PASS_FILE | string | `nil` | Path to the mapped file containing to the database password. (optional, replaces `DB_PASS`) |
--- a/charts/stable/wikijs/values.yaml
+++ b/charts/stable/wikijs/values.yaml
@@ -23,8 +23,10 @@ env:
   TZ: UTC
   # -- mysql, postgres, mariadb, mssql or sqlite
   DB_TYPE: sqlite
-  # -- Path to the SQLite file
+  # -- Path to the SQLite file if using sqlite
   DB_FILEPATH: /app/wiki/data/db.sqlite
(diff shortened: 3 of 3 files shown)


onedr0p:
I'm curious, if someone updates the user and group id in the environment variables won't they need to update fsGroup too? I'm not sure the best way to solve this other than putting in a note in the custom config readme and leaving podSecurityContext out of the values yaml.

bjw-s:
I'm also curious about the purpose of `app-wiki-data`, isn't this the same as `data` that we use in most of our other charts?

fezzik02:
So the issue here is that the `/data` folder which is used for a cache among other things is only writable by root in the generic install:
```sh
root@wikijs-generic-69c84f776b-68ljj:/# stat /data
  File: /data
  Size: 4096            Blocks: 8          IO Block: 4096   directory
Device: 5000dch/5243100d        Inode: 259065721   Links: 2
Access: (0755/drwxr-xr-x)  Uid: (    0/    root)   Gid: (    0/    root)
```
Here's the default user:
```sh
root@wikijs-generic-69c84f776b-68ljj:/# cat /etc/passwd | grep 911
abc:x:911:911::/config:/bin/false
```
```sh
root@wikijs-generic-69c84f776b-68ljj:/# ps aux | grep node
abc        288  1.1  0.1 320564 102772 ?       Ssl  17:54   0:03 /usr/bin/node server
root       363  0.0  0.0   1580   736 pts/1    S+   17:59   0:00 grep node
```
However, that user can't write to the `/data/cache/` path required:
```sh
root@wikijs-generic-69c84f776b-68ljj:/# su -s /bin/sh - abc
wikijs-generic-69c84f776b-68ljj:~$ whoami
abc
wikijs-generic-69c84f776b-68ljj:~$ mkdir /data/cache
mkdir: cannot create directory ‘/data/cache’: Permission denied
wikijs-generic-69c84f776b-68ljj:~$ 
```

When I provision PV's _and_ set the `fsGroup` to 911, folders are created as expected:
```sh
root@wikijs-7fb4f666c-njhjl:/# ls -la /data
total 19
drwxrwsr-x 5 root abc   1024 Aug 22 20:31 .
drwxr-xr-x 1 root root  4096 Aug 22 20:25 ..
drwxr-sr-x 2 abc  abc   1024 Aug 23 10:47 cache
drwxrws--- 2 root abc  12288 Aug 22 20:23 lost+found
drwxr-sr-x 2 abc  abc   1024 Aug 22 20:31 uploads
root@wikijs-7fb4f666c-njhjl:/# 
```
Meanwhile, `/app/wiki/data/` is the default home of sqlite, meaning yes there is different data in the two paths:
```sh
root@wikijs-7fb4f666c-njhjl:/# ls -la /app/wiki/data/
total 277
drwxrwsr-x 3 abc  abc    1024 Aug 23 20:31 .
drwxr-xr-x 1 root root   4096 Aug 22 20:25 ..
-rw-r--r-- 1 abc  abc  262144 Aug 23 20:31 db.sqlite
drwxrws--- 2 abc  abc   12288 Aug 22 20:23 lost+found
root@wikijs-7fb4f666c-njhjl:/# 
```

fezzik02:
Unfortunately, by themselves neither assigning a persistent drive nor setting the `fsGroup` resolve this issue.
```sh
2021-08-24T18:53:50.056Z [MASTER] info: Connecting to database...
2021-08-24T18:53:50.058Z [MASTER] error: Database Connection Error: SQLITE_CANTOPEN undefined:undefined
2021-08-24T18:53:50.058Z [MASTER] warn: Will retry in 3 seconds... [Attempt 8 of 10]
```

PR #1145, opened 2021-08-29
[games-on-whales] support for nvidia

angelnu:
**Description of the change**

**Benefits**

**Possible drawbacks**

**Applicable issues**

- fixes #

**Additional information**

**Checklist** 

### DIFF
--- a/charts/stable/games-on-whales/values.yaml
+++ b/charts/stable/games-on-whales/values.yaml
@@ -105,7 +105,7 @@ firefox:
     # -- image repository
     repository: andrewmackrodt/firefox-x11
     # -- image tag
-    tag: 1.0.0
+    tag: 91.0.2-r1
     # -- image pull policy
     pullPolicy: IfNotPresent
   # -- firefox log level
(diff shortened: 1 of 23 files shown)


charts/stable/games-on-whales/values.yaml
```diff
@@ -32,6 +32,10 @@ persistence:
     type: emptyDir
     mountPath: /home/retro
 
+# -- (object) Pass GPU resources to Xorg, steam and retroarch containers
+# See Custom configuration section in the Readme
+graphic_resources:
```
bjw-s, line 37: I'm curious what you see as the benefit of adding this key over the "regular" `resources` section since these resources are not really different. Is it because of the `additionalContainer`?

angelnu: it is because it is needed in multiple containers and it is tedious to add to the (right) 3 ones.

The option of copy&paste of the additionalContainer also works.

bjw-s (approved):

PR #1152, opened 2021-09-03
[speedtest-exporter] switch to serviceMonitor instead of podMonitor

jr0dd:
Since this chart deploys a service, I figured that it would be more appropriate for a serviceMonitor instead of a podMonitor.

**Checklist** 


onedr0p:
Looks good and makes sense!

onedr0p (approved):

PR #1153, opened 2021-09-04
[qbittorrent] add podMonitor and ability to add exporter as a sidecar

jr0dd:
The title is self explanatory lol

**Checklist** 


bjw-s:
Thanks for the PR! I have a few questions on this, but not enough time to write them out on mobile right now (almost bedtime 😴)

Will write them down here tomorrow!

bjw-s:
The thing I wanted to suggest is the following: we could have the chart automatically inject the additionalContainer if the podMonitor is enabled.

In order to do this you would have to switch to the "advanced" style of `common.yaml`, and do something similar to what e.g. the tdarr chart is doing: https://github.com/k8s-at-home/charts/blob/master/charts/stable/tdarr/templates/common.yaml.

That way the user doesn't have to think as much about enabling it. It would require a key in values.yaml where the image is specified (so that the user can override it / apps like renovate can bump it).

onedr0p:
Agreed! I'd like to see the arr exporters written that way too.

jr0dd:
> The thing I wanted to suggest is the following: we could have the chart automatically inject the additionalContainer if the podMonitor is enabled.
> 
> In order to do this you would have to switch to the "advanced" style of `common.yaml`, and do something similar to what e.g. the tdarr chart is doing: https://github.com/k8s-at-home/charts/blob/master/charts/stable/tdarr/templates/common.yaml.
> 
> That way the user doesn't have to think as much about enabling it. It would require a key in values.yaml where the image is specified (so that the user can override it / apps like renovate can bump it).

No problem. I'll get it worked up today. @onedr0p I can work up the pr's for the other 3 as well. I was already planning on adding that relabeling to those pod monitors. The one thing that irks me with podmonitors is that by default the job label has the namespace in front of it where service monitors do not.

onedr0p:
That would be awesome, as for the relabelings that fine too.

jr0dd:
@onedr0p @bjw-s if you guys think it needs any further adjusting just lmk before I start on the arr's

charts/stable/qbittorrent/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v2
-appVersion: v4.3.5
+appVersion: v4.3.7
 description: qBittorrent is a cross-platform free and open-source BitTorrent client
 name: qbittorrent
-version: 12.0.0
+version: 13.0.0
```
bjw-s, line 5: It looks like the changes are not breaking, so according to semver 12.1.0 would probably be better.

onedr0p:
@jr0dd do you think there is any benefit to include the additionalContainer with the podMonitor too? One method seems more than enough.

onedr0p:
oh wait, I have seen you remove that!

onedr0p (dismissed):

onedr0p:
@bjw-s this look good to you too?

charts/stable/qbittorrent/templates/podmonitor.yaml
```diff
@@ -0,0 +1,29 @@
+{{- if .Values.exporter.enabled }}
+apiVersion: monitoring.coreos.com/v1
+kind: PodMonitor
+metadata:
+  name: {{ template "common.names.fullname" . }}
+  labels:
+    {{- include "common.labels" . | nindent 4 }}
+    {{- with .Values.exporter.podMonitor.additionalLabels }}
```
bjw-s, line 8:
```suggestion
    {{- with .Values.exporter.podMonitor.labels }}
```

charts/stable/qbittorrent/values.yaml
```diff
@@ -60,3 +60,30 @@ persistence:
...
     mountPath: /downloads
+
+exporter:
+  # -- Enable and configure prometheus-qbittorrent-exporter sidecar and Prometheus podMonitor.
+  # @default -- See values.yaml
+  enabled: false
+  podMonitor:
+    interval: 15s
+    scrapeTimeout: 5s
+    additionalLabels: {}
```
bjw-s, line 71:
```suggestion
    labels: {}
```

bjw-s (changes requested):
Change looks good! Thanks for taking the time to make this 🍻 Just a few small changes requested :)

charts/stable/qbittorrent/Chart.yaml
jr0dd: ok

jr0dd:
@bjw-s alright I think everything is covered now

bjw-s (approved):

bjw-s:
Awesome! I'm gonna merge it as soon as the CI passes 👍

onedr0p:
@all-contributors add @jr0dd for code

PR #1155, opened 2021-09-05
[lidarr] refactor of exportarr sidecar and podmonitor

jr0dd:
If these changes look good I'll update sonarr and radarr as well.

**Checklist** 

### DIFF
--- a/charts/stable/lidarr/templates/common.yaml
+++ b/charts/stable/lidarr/templates/common.yaml
@@ -25,9 +25,11 @@ additionalContainers:
       - name: metrics
         containerPort: {{ .Values.exporter.env.port }}
     volumeMounts:
+      {{ if .Values.persistence.config.enabled }}
       - name: config
         mountPath: /config
         readOnly: true
+      {{ end }}
 {{ end }}
 {{- end -}}
 {{- $_ := mergeOverwrite .Values (include "lidarr.harcodedValues" . | fromYaml) -}}
--- a/charts/stable/lidarr/templates/common.yaml
+++ b/charts/stable/lidarr/templates/common.yaml
@@ -1,1 +1,34 @@
+{{/* Make sure all variables are set properly */}}
+{{- include "common.values.setup" . }}
+
+{{/* Append the hardcoded settings */}}
+{{- define "exportarr.harcodedValues" -}}
+{{ if .Values.exportarr.enabled }}
+additionalContainers:
+  exportarr:
+    name: exportarr
+    image: "{{ .Values.exportarr.image.repository }}:{{ .Values.exportarr.image.tag }}"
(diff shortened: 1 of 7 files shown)


onedr0p:
I think you can keep it consistent and make the key name `exporter` instead of `exportarr`

jr0dd:
> I think you can keep it consistent and make the key name `exporter` instead of `exportarr`, we can keep it like the qb one.

Yeah no problem. Are there any other additional env vars for exportarr you think should be included?

charts/stable/lidarr/values.yaml
```diff
@@ -66,35 +66,25 @@ persistence:
...
+    repository: ghcr.io/onedr0p/exportarr
+    # -- image tag
+    tag: v0.6.1
+    # -- image pull policy
+    pullPolicy: IfNotPresent
+  env:
+    # -- metrics port
+    port: 32123
+    # -- Set to true to enable gathering of additional metrics (slow)
+    addMetrics: false
```
onedr0p, line 88:
```suggestion
    additionalMetrics: false
```

charts/stable/lidarr/values.yaml
```diff
@@ -66,35 +66,25 @@ persistence:
...
+    # -- Set to true to enable gathering unknown queue items
+    unknownItems: false
```
onedr0p, line 90:
```suggestion
    unknownQueueItems: false
```

onedr0p:
I think you got all the ones that should be configurable. Just made a comment on to rename the 2 default ones

charts/stable/lidarr/templates/common.yaml
```diff
@@ -1 +1,36 @@
...
+        value: "{{ .Values.exporter.env.additionalMetrics }}"
+      - name: ENABLE_UNKNOWN_QUEUE_ITEMS
+        value: "{{ .Values.exporter.env.unknownQueueItems }}"
+    ports:
+      - name: metrics
+        containerPort: {{ .Values.exporter.env.port }}
+    volumeMounts:
+      - name: config
+        mountPath: /config
+        readOnly: true
```
bjw-s, line 30: This can potentially break when there is the `config` persistence item is disabled.

jr0dd: ok should be fixed

bjw-s (approved):

jr0dd:
@bjw-s or @onedr0p  So for whatever reason this didn't land in the repo. Sonarr and radarr went through just fine.

onedr0p:
I see in lidarr-12.0.0 in the github releases.

onedr0p:
If it's not coming thru we probably need to bump it to `12.0.1`.

PR #1170, opened 2021-09-09
[frigate] update default database path in configmap

### LINKED ISSUE #1109: Frigate chart version 5.0.0's value.yaml is confusing around persistence

onedr0p:
**Description of the change**

update default database path in configmap

**Benefits**

separate config data and stream data

**Possible drawbacks**

**Applicable issues**

- fixes #1109

**Additional information**

**Checklist** 

### DIFF
--- a/charts/stable/frigate/values.yaml
+++ b/charts/stable/frigate/values.yaml
@@ -99,8 +99,10 @@ resources: {}
 # @default -- see URL to default config
 config: |
   mqtt:
-    host: test.mosquitto.org
+    host: mosquitto
     topic_prefix: frigate
+  database:
+    path: /data/frigate.db
   detectors:
     cpu1:
       type: cpu
--- a/charts/stable/frigate/values.yaml
+++ b/charts/stable/frigate/values.yaml
@@ -99,7 +99,7 @@ resources: {}
 # @default -- see URL to default config
 config: |
   mqtt:
-    host: mosquitto
+    host: test.mosquitto.org
     topic_prefix: frigate
   database:
     path: /data/frigate.db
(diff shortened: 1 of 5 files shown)


charts/stable/frigate/values.yaml
```diff
@@ -99,8 +99,10 @@ resources: {}
 # @default -- see URL to default config
 config: |
   mqtt:
-    host: test.mosquitto.org
+    host: mosquitto
     topic_prefix: frigate
+  database:
+    path: /data/frigate.db
```
bjw-s, line 105: Will this work if there is no `/data` folder by default? It might make the CI / any first-time installs fail

onedr0p: Hmm this is true, should I make a note in the custom config section of the readme?

bjw-s: Yeah, that might be wise, and a `ct-values` file with an `emptyDir` is probably required.

bjw-s (approved):

PR #1268, opened 2021-11-06
[vikunja] Improve Vikunja chart.

LordGaav:
**Description of the change**

* Added a way to turn of Caddy's Let's Encrypt functionality, for those that use an Ingress controller that handles this already.
* Added a way to provide Vikunja with a YAML configuration file for options that can otherwise not be configured.

**Benefits**

* Better integration with Ingress controllers.
* Better configuration of Vikunja.

**Possible drawbacks**

None.

**Additional information**

**Checklist** 

### DIFF
--- a/charts/stable/vikunja/Chart.yaml
+++ b/charts/stable/vikunja/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 0.17.0
 description: The to-do app to organize your life
 name: vikunja
-version: 4.2.0
+version: 4.3.0
 keywords:
   - vikunja
   - to-do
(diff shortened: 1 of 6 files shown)


onedr0p:
@LordGaav Should we consider this a breaking change and bump the chart version accordingly?

LordGaav:
The only thing that might break for existing users is that the `jwtsecret` setting moved to the YAML file, and the YAML file is places in a location that a user might be placing a YAML already themselves (although I saw no way to do that, hence this Pull Request). The other changes are a Caddy flag that has the same default effect as before.

However, it might be the safe option, just to be clear that the new YAML configuration is new, and in my opinion the preferred way to configure Vikunja.

onedr0p (approved):

PR #1278, opened 2021-11-14
mealie: add mealie chart

roobre:
**Description of the change**

This commit adds a simple chart for Mealie, a self hosted recipe manager and meal planner.

**Checklist** 

### DIFF
--- a/charts/incubator/mealie/.helmignore
+++ b/charts/incubator/mealie/.helmignore
@@ -1,0 +1,26 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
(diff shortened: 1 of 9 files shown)


bjw-s:
Thanks for taking the time to create this PR!! I do believe it would be better to update our existing chart with any changes that you think are relevant 🙂 

https://github.com/k8s-at-home/charts/tree/master/charts/stable/mealie

roobre:
Thanks a lot for the heads up @bjw-s, I could swear I searched before bothering to create what was already there... But it seems like somehow I missed that 😅 

I'll take a quick look in case I can see anything worth adding, although this was a very basic first iteration so I don't think there's going to be anything.

Thanks and sorry for the noise!

PR #1299, opened 2021-12-08
[grocy] Upgrade to version 3.1.3 and fix upgrade bug

Lykos153:
**Description of the change**

- Added `emptyDir` volume for the viewcache in order to prevent problems when upgrading the image that would require manual intervention.
- Changed image tag to `version-v3.1.3`.

**Benefits**
No manual interaction required when upgrading the the image if persistence is enabled.
*grocy's upgrade instructions mention clearing `data/viewcache`. Failing to do so will result in an blank pages after upgrading the image. Mounting the directory as `emptyDir` removes the need to open a shell into the container and manually performing the operation after upgrading the image (which doesn't fit well in a k8s workflow).*

**Possible drawbacks**
After restarting the pod, the first time visiting the page might be a bit slower as the cache needs to be refilled.

**Checklist** 


bjw-s:
I have taken the liberty to update this PR so that it passes CI.

bjw-s (approved):

PR #1308, opened 2021-12-12
[tdarr] Add ffmpegPath environment override

npawelek:
**Description of the change**

This change adds and environment override for `ffmpegPath`. If unset, the
tdarr node container will use the default binary that ships with tdarr.
Using `""` may cause detection issues with hardware passthrough (`i915`),
but will remain backwards compatible with the container default.

**Benefits**

Allow overriding to the default system ffmpeg binary.

**Possible drawbacks**

None.

**Applicable issues**

n/a

**Additional information**

This issue was discussed at length with the container maintainer and is the best solution to override the default ffmpeg binary which has vaapi detection problems.

**Checklist** 

### DIFF
--- a/charts/stable/tdarr/Chart.yaml
+++ b/charts/stable/tdarr/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 2.00.10
 description: Tdarr is a self hosted web-app for automating media library transcode/remux management and making sure your files are exactly how you need them to be in terms of codecs/streams/containers etc.
 name: tdarr
-version: 4.1.3
+version: 4.1.4
 keywords:
   - transcoding
   - remux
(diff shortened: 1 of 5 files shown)


charts/stable/tdarr/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 2.00.10
 description: Tdarr is a self hosted web-app for automating media library transcode/remux management and making sure your files are exactly how you need them to be in terms of codecs/streams/containers etc.
 name: tdarr
-version: 4.1.3
+version: 4.1.4
```
bjw-s, line 5:
```suggestion
version: 4.2.0
```
This is not really a bugfix, but more of an added "feature", so a minor version bump is warranted.

bjw-s (changes requested):
Thanks for the ongoing work on this chart!! Just a small remark 🙂

npawelek:
@bjw-s No worries. Updated the readme to reflect the same as well.

bjw-s (approved):

PR #1309, opened 2021-12-13
[ghost] Add new chart

samip5:
**Description of the change**

Add new Chart for Ghost

**Benefits**

Able to deploy Ghost easilly

**Possible drawbacks**

The configuration of it, is far from optimal.

**Additional information**

**Checklist** 


charts/stable/ghost/README_CHANGELOG.md.gotmpl
```diff
@@ -0,0 +1,8 @@
+{{- define "custom.changelog.header" -}}
+## Changelog
+{{- end -}}
+
+{{- define "custom.changelog" -}}
+{{ template "custom.changelog.header" . }}
+
+{{- end -}}
```
bjw-s, line 8: Please use the original template for this file

charts/stable/ghost/values.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+# See https://ghost.org/docs/config/#running-ghost-with-config-env-variables
+env:
+  url: "http://some-ghost.example.com"
+  database__client: mysql
+  database__connection__host: mariadb
+  database__connection__user: ghost
+  database__connection__password: ghost
+  database__connection__database: ghost
+  NODE_ENV: production
```
bjw-s, line 24: Would it make sense to move this to the hardcoded section of the chart?

charts/stable/ghost/values.yaml
```diff
@@ -0,0 +1,58 @@
...
+  pullPolicy: IfNotPresent
+
+# See https://ghost.org/docs/config/#running-ghost-with-config-env-variables
+env:
+  url: "http://some-ghost.example.com"
+  database__client: mysql
+  database__connection__host: mariadb
+  database__connection__user: ghost
+  database__connection__password: ghost
+  database__connection__database: ghost
```
bjw-s, line 23: Since the app can be configured with sqlite, that might be a "better" default to provide. Then if users wish to use mysql, they can specify this themselves. Otherwise the chart will also not be able to pass the CI.

charts/stable/ghost/values.yaml
```diff
@@ -0,0 +1,58 @@
...
+    enabled: false
+
+# -- Configure persistence settings for the chart under this key.
+# @default -- See values.yaml
+persistence:
+  content:
+    enabled: false
+    mountPath: /var/lib/ghost/content
+
+
```
bjw-s, line 47:
```suggestion

```

bjw-s (changes requested):
Thanks for taking the time to create this PR! 🏅 I've left a few comments / questions.

charts/stable/ghost/values.yaml
samip5: No, it doesn't.

charts/stable/ghost/values.yaml
samip5: Provided that the storage in each own cluster can be pretty much anything, sqlite is heavily frown upon default, but I suppose that it could be changed to default to it.

charts/stable/ghost/values.yaml
bjw-s: Thank you for this elaborate comment. This really helps me understand the reasoning, and will get this PR approved só much quicker.

bjw-s: Apologies for the sass in my previous comment, I may be a bit grumpy form work-stuff 😊
I notice that the line selection for my comment isn't really presented well in the GitHub UI. My comment was only relating to the `NODE_ENV` variable.

charts/stable/ghost/values.yaml
bjw-s: It is also possible to stick with mysql, but then a different values file needs to be added for the CI to work. See for example the [Monica](https://github.com/k8s-at-home/charts/blob/master/charts/stable/monica/ci/ct-values.yaml) chart for an example

charts/stable/ghost/values.yaml
samip5: That portion does make sense to add there, but Ghost dislikes that production env if the db is sqlite so I don't know which approach is the better one here.

bjw-s: In that case I agree it's better to leave it configurable, since the other vars are configurable as well. 👍

samip5:
@bjw-s Could you try the CI again?

carpenike:
Closed and reopened, CI should kick off again.

bjw-s:
I forgot to reply that I approved the CI 😅 Looking at the logs, it seems like it is somehow trying to run with sqlite in the CI.

samip5:
> I forgot to reply that I approved the CI 😅 Looking at the logs, it seems like it is somehow trying to run with sqlite in the CI.

Yes, because that's what I told it to do.  Which apperantly doesn't work tho.

samip5:
@bjw-s & @carpenike Maybe now the CI will work? I did pretty much the same for it as is done for monica.

bjw-s:
Looks like the CI passed, I'll take a look at the chart in a bit 👍🏻

bjw-s (approved):
LGTM! 🥇

bjw-s:
@all-contributors Please add @samip5 for code

PR #1330, opened 2021-12-26
[cryptpad] Add chart (wip, rfc)

roobre:
**Description of the change**

This PR adds a chart for [Cryptpad](https://cryptpad.fr/), a collaborative e2e encrypted suite.

**Possible drawbacks**

This is a WIP/RFC (request for comments) PR. While I have this running nicely in my cluster, this particular application is a bit bulky and I had to make some non obvious choices to make it work. Feedback and suggestions on how to improve this are more than welcome!

**Applicable issues**

**Additional information**

**Checklist** 


charts/incubator/cryptpad/templates/common.yaml
```diff
@@ -0,0 +1,16 @@
...
+mountPath: "/cryptpad/config/config.js"
+subPath: "config.js"
+type: "custom"
+volumeSpec:
+  configMap:
+    name: {{ template "common.names.fullname" . }}-config
+{{- end -}}
+{{- $_ := set .Values.persistence "cryptpad-config" (include "cryptpad.configVolume" . | fromYaml) -}}
+
+{{ include "common.all" . }}
```
roobre, line 16: I am not completely sure if this is idiomatic, but since I did have to use a configMap for this application to work I copy-pasted this from https://github.com/k8s-at-home/charts/blob/master/charts/stable/mosquitto/templates/common.yaml.

charts/incubator/cryptpad/templates/configmap.yaml
```diff
@@ -0,0 +1,23 @@
...
+data:
+  config.js: |
+    module.exports = {
+      {{- if and .Values.ingress.main.enabled .Values.ingress.main.tls }}
+      httpUnsafeOrigin: 'https://{{ (index (.Values.ingress.main.tls | first).hosts 0) }}',
+      httpSafeOrigin:   'https://{{ (index (.Values.ingress.main.tls | first).hosts 1) }}',
+      {{- else if .Values.ingress.main.enabled }}
+      httpUnsafeOrigin: '{{ (index .Values.ingress.main.hosts 0).host }}',
+      httpSafeOrigin:   '{{ (index .Values.ingress.main.hosts 1).host }}',
+      {{- end }}
```
roobre, line 16: Cryptpad requires these variables to work. I chose to be "smart" here and just generate this from the two ingressess in the values (if present) to keep config simple, but perhaps you see this too convoluted. Feedback welcome!

charts/incubator/cryptpad/values.yaml
```diff
@@ -0,0 +1,78 @@
...
+# -- environment variables. See more environment variables in the [cryptpad documentation](https://cryptpad.org/docs).
+# @default -- See below
+env:
+  # -- Set the container timezone
+  TZ: UTC
+  # Main and sandbox domains are automatically fetched from the first and second host defined in the ingress,
+  # respectively.
+  # These env vars are used in the internal nginx.
+  CPAD_MAIN_DOMAIN: "{{ (index .Values.ingress.main.hosts 0).host }}"
+  CPAD_SANDBOX_DOMAIN: "{{ (index .Values.ingress.main.hosts 1).host }}"
```
roobre, line 38: This is again the same thing as the configmap: Attempt to be smart and not make the user input the same thing in two places.

Settings appear twice (as env vars and in the CM) because there are actually two different components reading these: The application itself, and an entrypoint script that generates an internal nginx config.

charts/incubator/cryptpad/Chart.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: v2
+appVersion: v4.12.0-nginx
```
roobre, line 2: Cryptpad provides images with and without a built-in nginx reverse proxy.

In my first approach I attempted to use the one without the built-in  nginx, to avoid having an "extra" reverse proxy behind the ingress controller. However, this turned out to be very hard: Cryptpad requires a very specific config with two domains and a bunch of different endpoints for static data and different rules for setting headers.

Having two reverse proxies is not the best but it is definitely more maintainable.

onedr0p:
I am ok with this getting merged into incubator if you still want to hack at it to make it better!.

charts/incubator/cryptpad/templates/common.yaml
onedr0p: This looks fine to me :D

charts/incubator/cryptpad/templates/configmap.yaml
onedr0p: Also looks fine, good idea to be honest

charts/incubator/cryptpad/Chart.yaml
onedr0p: This is pretty much like any other application in which a programming language doesn't have a built in web server. Using nginx is fine with me

onedr0p:
@all-contributors add @roobre  for code

roobre:
Thanks for the feedback @onedr0p!

I'll keep toying with it and sending PRs whenever I see it breaks :)

PR #1337, opened 2021-12-30
[blocky] use container image direct from the application developer

haraldkoch:
It's possible that when this chart was first built, container images (or multi-arch images) were not available - but they are now. See https://github.com/0xERR0R/blocky/pkgs/container/blocky for images that include amd64, arm64, arm/v6, and arm/v7.

**Description of the change**

This change changes the blocky helm chart to use container images built by the app maintainer.

**Benefits**

Container images direct from the source are published sooner and theoretically more reliable.

**Possible drawbacks**

unknown?

**Applicable issues**

- fixes #

**Additional information**

**Checklist** 

### DIFF
--- a/charts/stable/blocky/Chart.yaml
+++ b/charts/stable/blocky/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: v0.17
 description: DNS proxy as ad-blocker for local network
 name: blocky
-version: 10.0.0
+version: 10.0.1
 kubeVersion: ">=1.16.0-0"
 keywords:
 - blocky
--- a/charts/stable/blocky/README.md
+++ b/charts/stable/blocky/README.md
@@ -1,6 +1,6 @@
 # blocky
 
-![Version: 10.0.0](https://img.shields.io/badge/Version-10.0.0-informational?style=flat-square) ![AppVersion: v0.17](https://img.shields.io/badge/AppVersion-v0.17-informational?style=flat-square)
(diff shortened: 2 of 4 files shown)


haraldkoch:
I updated the Chart.yaml version after I first created the pull request. Since the two container images should be effectively identical, I didn't think a semver update beyond "patch" was required, but I can change that.

It seems changelog entries are only added for major version changes, and this is a minor update?

bjw-s:
> I updated the Chart.yaml version after I first created the pull request. Since the two container images should be effectively identical, I didn't think a semver update beyond "patch" was required, but I can change that.

Updating the patch level is fine 👍

As for the changelog, you will need to modify the `README_CHANGELOG.md.gotmpl` file, since the README.md file is automatically generated from those templates and your current changes would be overwritten.

> It seems changelog entries are only added for major version changes, and this is a patch update?

We normally add change notes to all releases except for the ones that are because of minor/patch version updates of the underlying common library (the tool we use for the updates has no idea about our changelogs, and it would be a lot of manual work to update all the chart docs).

haraldkoch:
Got it. None of the charts I use have had minor updates that I've seen, I incorrectly assumed this was a pattern.

Does that look better?

onedr0p (approved):

haraldkoch:
Two questions:

1) "This branch is out-of-date" - rebase from master, or just merge the changes?
2) should I squash the four commits into one (and force-push the branch)? Or just leave it alone?

onedr0p:
You are all good, unless there were conflicts we just squash and merge (even if the branch is technically out of date)

Issue #1358, opened 2022-01-05
[ghost] Missing PVC


Brice187:
# Details

**Helm chart name and version:**

kustomization.yaml:
```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

helmCharts:
- name: ghost
  version: 1.1.0
  repo: https://k8s-at-home.com/charts/
  valuesFile: values.yaml
  namespace: ghost
  releaseName: ghost
```

Running `kustomize build --enable-helm` creates following yaml(-snippet):
```yaml
...
---
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app.kubernetes.io/instance: ghost
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: ghost
    app.kubernetes.io/version: 4.27.2
    helm.sh/chart: ghost-1.1.0
  name: ghost
spec:
  replicas: 1
  revisionHistoryLimit: 3
  selector:
    matchLabels:
      app.kubernetes.io/instance: ghost
      app.kubernetes.io/name: ghost
  strategy:
    type: Recreate
  template:
    metadata:
      labels:
        app.kubernetes.io/instance: ghost
        app.kubernetes.io/name: ghost
    spec:
      automountServiceAccountToken: true
      containers:
      - env:
        - name: NODE_ENV
          value: production
        - name: database__client
          value: mysql
        - name: database__connection__database
          value: ghost
        - name: database__connection__host
          value: ghost-mariadb
        - name: database__connection__password
          value: XXXXXXXXXXXXX_
        - name: database__connection__user
          value: ghost
        - name: url
          value: http://ghost.local
        image: ghost:4.27.2
        imagePullPolicy: IfNotPresent
        livenessProbe:
          failureThreshold: 3
          initialDelaySeconds: 0
          periodSeconds: 10
          tcpSocket:
            port: 2368
          timeoutSeconds: 1
        name: ghost
        ports:
        - containerPort: 2368
          name: http
          protocol: TCP
        readinessProbe:
          failureThreshold: 3
          initialDelaySeconds: 0
          periodSeconds: 10
          tcpSocket:
            port: 2368
          timeoutSeconds: 1
        startupProbe:
          failureThreshold: 30
          initialDelaySeconds: 0
          periodSeconds: 5
          tcpSocket:
            port: 2368
          timeoutSeconds: 1
      dnsPolicy: ClusterFirst
      enableServiceLinks: true
[...]

PR #1359, opened 2022-01-06
[ghost] Fix PVC creation

bjw-s:
Signed-off-by: Bᴇʀɴᴅ Sᴄʜᴏʀɢᴇʀs <me@bjw-s.dev>

**Description of the change**

**Benefits**

**Possible drawbacks**

**Applicable issues**

- fixes #1358 

**Additional information**

**Checklist** 
- [ ] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [ ] Title of the PR starts with chart name (e.g. `[home-assistant]`)
- [ ] Variables are documented in the README.md (this can be done with using our helm-docs wrapper `./hack/gen-helm-docs.sh stable <chart>`)

PR #1360, opened 2022-01-06
[ghost] Fix PVC creation

lcotonea:
The current common file is not enough if the .Value contains persistence item is different than "content".
For exemple if you specify these following values, the persistence item shared will not be processed:
```YAML
persistence:
  content:
    enabled: false
    existingClaim: "pvc-nfs-ghost"
    mountPath: "/var/lib/ghost/content"
  shared:
    enabled: true
    mountPath: /shared
    type: emptyDir
```

**Description of the change**

All persistent items are computed in the hardcoded settings.

**Benefits**

We can specify none, only content, or several persistent items.

**Possible drawbacks**

None

**Applicable issues**

- fixes [ghost] Missing PVC #1358

**Additional information**

**Checklist** 

### DIFF
--- a/charts/stable/ghost/templates/common.yaml
+++ b/charts/stable/ghost/templates/common.yaml
@@ -3,13 +3,22 @@
 
 {{/* Append the hardcoded settings */}}
 {{- define "ghost.hardcodedValues" -}}
-  {{- if not .Values.persistence.content.enabled }}
+
 persistence:
+  {{- range $index, $persistence := .Values.persistence }}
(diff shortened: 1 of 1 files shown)


bjw-s:
As stated in the original PR, I am not a big fan of iterating over the existing values. 

I tried installing the chart without enabling the `content` persistence item, and that seems to work fine. What would be the use case of setting this to an emptyDir if the item is not enabled?

bjw-s:
Closed as discussed in #1359

PR #1372, opened 2022-01-12
[thelounge] Update image version to 4.3.0-alpine from 4.2.0-alpine

psych0d0g:
just bump a version, nothing less nothing more ;)


samip5:
It just needs an Chart version number update too and it would pass the lint.

PR #1375, opened 2022-01-16
[helm-docs] Update reference to k8s-at-home specific helm-docs

TheDJVG:
**Description of the change**

This project requires a custom helm-docs version. Make sure we [refer to this version](https://github.com/k8s-at-home/helm-docs) when it's not installed.


bjw-s:
Good catch!

PR #1403, opened 2022-02-06
[teamspeak] Add TeamSpeak metrics.

reitermarkus:
**Description of the change**

- Merge ports into single service (needs `MixedProtocolLBService` feature). Currently only the `unifi` chart seems to mention and use this feature.
- Add `ts3exporter` metrics sidecar.

**Benefits**

- In my case, no need for MetalLB's `allow-shared-ip` annotation anymore.
- Metrics.

**Possible drawbacks**

N/A

**Applicable issues**

N/A

**Additional information**

N/A

**Checklist** 
- [x] Title of the PR starts with chart name (e.g. `[home-assistant]`)
- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [x] Chart `artifacthub.io/changes` changelog annotation has been updated in `Chart.yaml`. See [Artifact Hub documentation](https://artifacthub.io/docs/topics/annotations/helm/#supported-annotations) for more info.
- [x] Variables have been documented in the `values.yaml` file.

PR #1404, opened 2022-02-06
[traefik-forward-auth] Reuse existing secret.

reitermarkus:
**Description of the change**

Don't generate a new secret on every upgrade.

**Benefits**

Does not result in “Not authorized” errors after an upgrade.

**Possible drawbacks**

N/A

**Applicable issues**

N/A

**Additional information**

**Checklist** 

---

Closes https://github.com/k8s-at-home/charts/pull/1406.


bjw-s:
Hi! Thanks for creating this PR! 

I'm wondering if instead it would be possible to provide the user with an option to specify an `existingSecret`? That seems to be an approach that I've seen in other charts as well. That way we don't rely on the naming logic, and the behavior would be more explicit.

So if `.Values.existingSecret` is set, that is _always_ used and no new Secret object is created. 
If `.Values.secret` is set, the Secret object is created/managed by Helm with the specified value, and if neither is set a new random secret is generated.

reitermarkus:
Currently, it is possible to set either `secret` or `cookie.secret` or both, so this wouldn't quite map easily to a single `existingSecret`.

We could do the same thing as I did in https://github.com/k8s-at-home/charts/pull/1403 with `env.TS3SERVER_SERVERADMIN_PASSWORD` and check if `env.SECRET` or `env.COOKIE_SECRET` is already set.

This way a user can set the secret directly with `env.SECRET` or use e.g. `env.SECRET.secretKeyRef`.

bjw-s:
> Currently, it is possible to set either `secret` or `cookie.secret` or both, so this wouldn't quite map easily to a single `existingSecret`.
> 
> We could do the same thing as I did in #1403 with `env.TS3SERVER_SERVERADMIN_PASSWORD` and check if `env.SECRET` or `env.COOKIE_SECRET` is already set.
> 
> This way a user can set the secret directly with `env.SECRET` or use e.g. `env.SECRET.secretKeyRef`.

That's a good point, I didn't think of that. Sounds like a good idea to approach in a similar way as #1403 :+1:

reitermarkus:
Okay, I changed this now to depend on the `common` chart and refactored it to only use `env` instead of duplicating all variables in `values.yaml`.

charts/stable/traefik-forward-auth/Chart.yaml
```diff
@@ -17,3 +17,13 @@ sources:
...
   - name: DirtyCajunRice
     email: nick@cajun.pro
+dependencies:
+  - name: common
+    repository: https://library-charts.k8s-at-home.com
+    version: 4.3.0
+annotations:
+  artifacthub.io/changes: |
+    - kind: added
+      description: Added `common` chart dependency.
```
bjw-s, line 27:
```suggestion
      description: **Breaking** Added `common` chart dependency. This will (potentially) require you to update your chart values.
```

charts/stable/traefik-forward-auth/values.yaml
```diff
@@ -1,190 +1,68 @@
...
+# This chart inherits from our common library chart. You can check the default values/options here:
+# https://github.com/k8s-at-home/library-charts/tree/main/charts/stable/common/values.yaml
+#
 
 image:
+  # -- image repository
   repository: thomseddon/traefik-forward-auth
-  pullPolicy: IfNotPresent
+  # -- image tag
   tag: ""
```
bjw-s, line 12:
```suggestion
  # -- image tag
  # @default -- chart.appVersion
  tag:
```

bjw-s (changes requested):

charts/stable/traefik-forward-auth/ci/ct-values.yaml
```diff
@@ -1,5 +1,3 @@
-providers:
-  google:
-    enabled: true
-    clientId: "fakeclientid"
-    clientSecret: "fakeclientsecret"
+env:
+  PROVIDERS_GOOGLE_CLIENT_ID: fakeclientid
+  PROVIDERS_GOOGLE_CLIENT_SECRET: fakeclientsecret
```
bjw-s, line 3:
```suggestion
  PROVIDERS_GOOGLE_CLIENT_SECRET: fakeclientsecret

middleware:
  enabled: false
```

It makes sense to enable the middleware by default, but in our CI environment we don't have that CRD.

bjw-s (changes requested):

bjw-s (approved):

PR #1410, opened 2022-02-11
[dendrite] Bump dendrite app version to v0.6.3

Jonnobrow:
**Description of the change**

Updates the app version to v0.6.3 from v0.6.2

**Benefits**

v0.6.2 contained a number of issues which are fixed by v0.6.3. Updating the default app version will give people using the chart a better out-of-the-box experience.

**Possible drawbacks**

N/A

**Applicable issues**

N/A

**Additional information**

**Checklist** 

### DIFF
--- a/charts/incubator/dendrite/values.yaml
+++ b/charts/incubator/dendrite/values.yaml
@@ -10,7 +10,7 @@ image:
   # -- image repository
   repository: matrixdotorg/dendrite-monolith
   # -- image tag
-  tag: "v0.6.2"
+  tag: "v0.6.3"
   # -- image pull policy
   pullPolicy: IfNotPresent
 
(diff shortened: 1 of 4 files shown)


charts/incubator/dendrite/values.yaml
```diff
@@ -10,7 +10,7 @@ image:
   # -- image repository
   repository: matrixdotorg/dendrite-monolith
   # -- image tag
-  tag: "v0.6.2"
+  tag: "v0.6.3"
```
bjw-s, line 13: Could it make sense to replace all of these with the following snippet?

```yaml
  # -- image tag
  tag:
```

That way, the version gets read from the chart metadata by default, and it saves us having to specify it so many times.

bjw-s (changes requested):
Thanks for submitting the PR!! It looks good to me, I just have a small change request (if it doesn't introduce too much trouble).

charts/incubator/dendrite/values.yaml
Jonnobrow: Yeah I agree that is much nicer. Because of the way I am using Subcharts, there was a little extra work to pass the app version through to those but nothing too major :)

bjw-s (approved):

PR #1412, opened 2022-02-13
[seafile] Add warning about naming collision

ellakk:
**Description of the change**

Add a note about a naming collision that will occur if you use **shared** as a key when configuring persistence.
I got bit by this myself. Instead of getting a PVC by my default storage class, I got an emptyDir. I did not notice this and when I restarted my node I lost all my data. I had another backup, so nothing valuable was lost, but I figure a warning would be nice so other people don't make the same mistake.

**Benefits**
Avoid mistakes and confusion when configuration the chart, especially if you are new to the K8S@HOME charts.

**Checklist** 


bjw-s:
`shared` isn't reserved per se. However, if only `enabled: true` is set, the other values will come from the underlying library chart.
In other words, if you (for example) specify the rest of the values as follows, it could be used just like any other persistence item:

```yaml
persistence:
  shared:
    enabled: true
    type: pvc
    accessMode: ReadWriteOnce
    size: 1Gi
```

Looking at the chart, I am wondering what the use case is for the `/shared` folder. I don't really see it being used anywhere (by default), and therefore I think it could be removed from the `values.yaml` completely.

Additionally, since this chart is now on v2.x already, could you please move it into the `stable` folder? We want to keep `incubator` for charts that are in kind of a "beta" phase.

ellakk:
That is true, but a warning is still good, because Seafile by default use a folder named "shared" in the root of the container for state. The reason why I did not add "type:pvc" to my values.yml is because on any other key you will get pvc by default if left empty, I did not realize that "shared" was preconfigured to use emptyDir. I think the example should still be in values to make it clear  that if you want to save application state, a mount is needed on /shared, this seem to be standard on most other charts.

Do you want a new PR for the move to stable?

bjw-s:
> That is true, but a warning is still good, because Seafile by default use a folder named "shared" in the root of the container for state. The reason why I did not add "type:pvc" to my values.yml is because on any other key you will get pvc by default if left empty, I did not realize that "shared" was preconfigured to use emptyDir. I think the example should still be in values to make it clear that if you want to save application state, a mount is needed on /shared, this seem to be standard on most other charts.

That makes sense 👍 

> Do you want a new PR for the move to stable?

Ideally it is done in this PR, since the chart-releaser CI will then not have to recreate a release 🙂 Thanks!

ellakk:
I've moved the chart, but i''m not sure what causes the "validate changelog" task to fail.

bjw-s:
That may be due to the validation script. That doesn't take the scenario of a chart moving from incubator > stable in to account 😊 I'll check that tomorrow to be sure and will merge 👍

bjw-s (approved):

PR #1418, opened 2022-02-19
[rtsp-to-web] Add initial revision of RTSPtoWeb helm chart

allenporter:
**Description of the change**

This is the initial version of a helm chart for RTSPtoWeb, a go binary that can be used to convert RTSP streams to WebRTC, HLS, LL HLS, MSE.   See https://github.com/deepch/RTSPtoWeb for the original project and https://github.com/deepch/RTSPtoWeb/issues/91 for context on the docker image created.

**Benefits**

This chart makes it easier to use with the Home Assistant integration https://www.home-assistant.io/integrations/rtsp_to_webrtc/ which requires running an external server, similar to zwavejs. This is an alternative to running as a Home Assistant Add-on e.g. https://community.home-assistant.io/t/add-on-rtsptoweb-and-rtsptowebrtc/387846

**Possible drawbacks**

The docker image itself or chart could likely be improved to make it easier to customize the /app/config.json file, however, you can also use this service just using the API without adjusting the configuration file. Additionally, it uses the host network given it has a bunch of UDP traffic for video streams.

**Applicable issues**

- fixes #

**Additional information**

**Checklist** 


allenporter:
This is an example of what it looks like to deploy this in k8s without using a helm chart at the moment:
https://github.com/allenporter/k8s-gitops/blob/main/home/dev/stream/rtsp-to-web.yaml

Perhaps you can see if i'm missing anything obvious that needs to go into the template. I just kind of compared to home assistant which uses host network and it seems like I don't need anything else.  From my local testing of this chart it seemed to work, but I may have missed something.

onedr0p (approved):

PR #1426, opened 2022-02-21
[strongswan] Add StrongSwan chart.

reitermarkus:
**Description of the change**

Add chart for my [StrongSwan docker image](https://github.com/reitermarkus/strongswan) which automatically generates a `.p12` certificate and `.mobileconfig` configuration.

**Benefits**

Having a chart.

**Possible drawbacks**

Very opinionated image.

**Applicable issues**

N/A

**Additional information**

**Checklist** 


bjw-s:
It looks like the install phase is missing a `VPN_DOMAIN` environment variable, which is causing the CI to fail.

charts/incubator/strongswan/values.yaml
```diff
@@ -0,0 +1,74 @@
...
+
+# -- Configures probes for the chart. Normally this does not need to be modified.
+# @default -- Disabled
+probes:
+  liveness:
+    enabled: false
+  readiness:
+    enabled: false
+  startup:
+    enabled: false
```
bjw-s, line 67: Should the probes always be disabled or should it be possible for the user to enable them?

reitermarkus: There is no HTTP service to probe, so it should always be disabled. I'm not sure if there is another way to probe StrongSwan.

bjw-s: No problem... In that case, something like this https://github.com/k8s-at-home/charts/blob/master/charts/stable/modem-stats/templates/common.yaml#L22L29 could be even better, since it prevents users from "accidentally" enabling the probes somehow.

reitermarkus: Okay, I found a simple command which should work as a rudimentary probe.

charts/incubator/strongswan/values.yaml
```diff
@@ -0,0 +1,74 @@
...
+#
+# This chart inherits from our common library chart. You can check the default values/options here:
+# https://github.com/k8s-at-home/library-charts/tree/main/charts/stable/common/values.yaml
+#
+
+image:
+  # -- image repository
+  repository: ghcr.io/reitermarkus/strongswan
+  # -- image tag
+  tag: latest
```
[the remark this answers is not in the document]
reitermarkus: Okay, I tagged a release.

bjw-s (approved):

PR #1428, opened 2022-02-21
[rtsp-to-web] Update RTSPtoWeb to 2.2.0 and update config location

allenporter:
**Description of the change**

Update RTSPtoWeb to 2.2.0 which is the latest version that now has images for multiple platforms (arm, etc).  Updated the persistence examples to work with a more standard location under a `/config` volume. Remove outdated comments.

**Benefits**

The updated configuration example should better describe that there is a `/config` volume.

**Possible drawbacks**

An alternative could be to include a full `config.json` more directly in the chart, but it was not clear this is worthwhile yet.

**Applicable issues**

- fixes #

**Additional information**

**Checklist** 

### DIFF
--- a/charts/stable/rtsp-to-web/Chart.yaml
+++ b/charts/stable/rtsp-to-web/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 1.0.0
 description: RTSPtoWeb converts RTSP streams to WebRTC, LL-HLS, HLS, MSE
 name: rtsp-to-web
-version: 1.0.0
+version: 2.0.0
 kubeVersion: ">=1.16.0-0"
 keywords:
 - rtsp-to-web
@@ -19,5 +19,7 @@ dependencies:
   version: 4.3.0
 annotations:
   artifacthub.io/changes: |
-    - kind: added
-      description: Initial version
+    - kind: changed
+      description: Upgrade RTSPtoWeb to v2.2.0
+    - kind: changed
+      description: The config directory is now located at `/config/config.json`
--- a/charts/stable/rtsp-to-web/Chart.yaml
+++ b/charts/stable/rtsp-to-web/Chart.yaml
@@ -22,4 +22,4 @@ annotations:
     - kind: changed
       description: Upgrade RTSPtoWeb to v2.2.0
     - kind: changed
-      description: The config directory is now located at `/config/config.json`
+      description: **BREAKING** The config directory is now located at `/config/config.json`
--- a/charts/stable/rtsp-to-web/Chart.yaml
+++ b/charts/stable/rtsp-to-web/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v2
-appVersion: 1.0.0
+appVersion: v2.2.0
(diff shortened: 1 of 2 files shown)


charts/stable/rtsp-to-web/values.yaml
```diff
@@ -36,9 +36,10 @@ ingress:
 # -- Enable WebRTC and other video protocols
 # hostNetwork: true
 
-# -- Configure persistence settings for the chart under this key.
+# -- Configure persistence settings for the chart under this key. See https://github.com/deepch/RTSPtoWeb#configuration
+# for /config/config.json settings and examples.
```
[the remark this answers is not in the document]
allenporter: So it actually will write back to config.json if you add or remove streams in the interface. So it's possible to use it with a static config you supply or it's possible to persist it.  The config file is the only interesting file.

I currently use this with a configmap, but others seem like they want to modify e.g. https://github.com/deepch/RTSPtoWeb/issues/107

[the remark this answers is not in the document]
allenporter: Are there charts that do both? I think I lining the config file would likely be more common.

charts/stable/rtsp-to-web/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 1.0.0
```
onedr0p, line 2:
```suggestion
appVersion: v2.2.0
```

charts/stable/rtsp-to-web/values.yaml
```diff
@@ -9,11 +9,11 @@ image:
   # -- image repository
   repository: ghcr.io/deepch/rtsptoweb
   # -- image tag
-  tag: v1.0.0
+  tag: v2.2.0
```
onedr0p, line 12:
```suggestion
  tag:
```

onedr0p: This will default to the `appVersion` in the `Chart.yaml`. Allows us to only maintain the container version in one place.

onedr0p (changes requested):
Couple comments

charts/stable/rtsp-to-web/values.yaml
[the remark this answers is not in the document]
bjw-s: If you want to have it both ways you'd need to introduce an initContainer that selectively copies over the configMap to the target volume. But that seems a bit heavy, so keeping it like this is fine with me if it works :)

charts/stable/rtsp-to-web/Chart.yaml
allenporter: I lost these when resolving merge conflicts and re-applied manually.

allenporter:
Do you squash & merge when accepting the PR, or prefer to squash commits? (or none)

bjw-s:
We normally try to squash & merge, and clean up the generated commit message manually (at least when it lets me, since the iOS app doesn't)

onedr0p:
@allenporter @bjw-s you all set here?

allenporter:
Yes, ready for review

onedr0p (approved):

PR #1429, opened 2022-02-21
Update contributing instructions for breaking changes

allenporter:
**Description of the change**

Update documentation for how to update breaking changes based on the move to annotations and artifacthub.

**Benefits**

**Possible drawbacks**

**Applicable issues**

- fixes #

**Additional information**

**Checklist** 
- [ ] Chart `artifacthub.io/changes` changelog annotation has been updated in `Chart.yaml`. See [Artifact Hub documentation](https://artifacthub.io/docs/topics/annotations/helm/#supported-annotations) for more info.
- [ ] Variables have been documented in the `values.yaml` file.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -47,10 +47,8 @@ Chart releases must be immutable. Any change to a chart warrants a chart version
 
 The chart `version` should follow [semver](https://semver.org/).
 
-Charts should start at `1.0.0`. All changes should be described in the [Chart metadata](https://docs.k8s-at-home.com/our-helm-charts/development/creating-a-new-chart/#chart-metadata). Additionally, any breaking (backwards incompatible) changes to a chart should: 
-
-1. Bump the MAJOR version
-1. Include `**BREAKING**` in the change description and describe the manual steps necessary to upgrade to the new (specified) MAJOR version
+Charts should start at `1.0.0`. Any any breaking (backwards incompatible) changes should Bump the MAJOR version, and should describe the manual steps necessary to upgrade.  All changes should be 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -47,10 +47,10 @@ Chart releases must be immutable. Any change to a chart warrants a chart version
 
 The chart `version` should follow [semver](https://semver.org/).
 
-Charts should start at `1.0.0`. Any breaking (backwards incompatible) changes to a chart should:
+Charts should start at `1.0.0`. All changes should be described in the [Chart metadata](https://docs.k8s-at-home.com/our-helm-charts/development/creating-a-new-chart/#chart-metadata). Additionally, any breaking (backwards incompatible) changes to a chart should: 
 
 1. Bump the MAJOR version
-2. In the README, under a section called "Upgrading", describe the manual steps necessary to upgrade to the new (specified) MAJOR version
+1. Include `**BREAKING**` in the change description and describe the manual steps necessary to upgrade to the new (specified) MAJOR version
```
bjw-s, line 53: I know some of this wording was already in place, but maybe we should change it a bit. I am not sure we really want extensive upgrade instructions, but I do believe that any breaking changes should be  communicated. I'm not sure what the correct balance would be though 😅

allenporter: Yes, I think the main motivation is just to remove what is here which I think is incorrect. So having less detail makes sense as it would be less to maintain.

How about this?  I've tried to leave the spirit of the message here (to tell people that major versions are important, and it is important to describe upgrade steps) but still linked to the more detailed chart instructions that describe the annotation metadata.

bjw-s (approved):

bjw-s:
Thanks!! LGTM :)

PR #1435, opened 2022-02-23
[pod-gateway] Permit overriding DNSPolicy

funkypenguin:
**Description of the change**

Hey gang! pod-gateway is a work of genius! This PR introduces a small optional changes to allow the operator to alter the `DNSPolicy` on the mutated pod.

Why? In my use-case, I mutate pods which run initContainers, and these initContainers require working DNS / network connectivity. Without this change, the webhook will always apply `DNSPolicy: None` to the pod, and then explicitly specify the VXLAN gateway endpoint as the DNS server using `dnsConfig`. This only works when the sidecar is up though, and so breaks DNS resolution for initContainers.

If a user doesn't care about routing DNS traffic over the gateway, they can set `DNSPolicy: ClusterFirst`, which will effectively ignore the `dnsConfig` values, and cause the pod to use the clusterDNS (_which is Kubernetes' default pod behavior_)

**Benefits**

Users who (a) need initContainers to have DNS and (b) don't care about routing this DNS traffic over the gateway can now do so.

**Possible drawbacks**

None

**Additional information**

**Checklist** 

### DIFF
--- a/charts/stable/pod-gateway/templates/webhook-deployment.yaml
+++ b/charts/stable/pod-gateway/templates/webhook-deployment.yaml
@@ -50,7 +50,7 @@ spec:
             # Static
             - --tls-cert-file-path=/tls/tls.crt
             - --tls-key-file-path=/tls/tls.key
-            - --DNSPolicy=DNS={{ .Values.DNSPolicy }}
+            - --DNSPolicy={{ .Values.DNSPolicy }}
             # Init container
             - --initImage={{ .Values.image.repository }}:{{ .Values.image.tag }}
             - --initImagePullPol={{ .Values.image.pullPolicy }}
--- a/charts/stable/pod-gateway/templates/webhook-deployment.yaml
+++ b/charts/stable/pod-gateway/templates/webhook-deployment.yaml
@@ -50,7 +50,7 @@ spec:
             # Static
             - --tls-cert-file-path=/tls/tls.crt
             - --tls-key-file-path=/tls/tls.key
-            - --DNSPolicy=None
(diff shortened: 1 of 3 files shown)


onedr0p:
@angelnu what do you think of this change?

angelnu:
Thanks for the change @funkypenguin - I could not think of a usage for the option so I did not add a config for it in the chart. But your use case does make sense - good explanation!

angelnu (dismissed):

angelnu:
The a couple of trailing spaces to remove in the `values.yaml` comments - could you please fix those @funkypenguin so we can merge? Thanks in advance.

funkypenguin:
Fast response guys! :) I've zapped the two little invisible buggers! 👾

charts/stable/pod-gateway/templates/webhook-deployment.yaml
```diff
@@ -50,7 +50,7 @@ spec:
             # Static
             - --tls-cert-file-path=/tls/tls.crt
             - --tls-key-file-path=/tls/tls.key
-            - --DNSPolicy=None
+            - --DNSPolicy=DNS={{ .Values.DNSPolicy }}
```
onedr0p, line 53: Shouldn't this be `- --DNSPolicy={{ .Values.DNSPolicy }}` ?

funkypenguin: Oops. Totally. That was a failed paste! Fixed now!

onedr0p (approved):

PR #1444, opened 2022-03-06
[sdtd] Add 7 Days to Die chart.

reitermarkus:
**Description of the change**

Add chart for 7 Days to Die server.

**Benefits**

Having a chart.

**Possible drawbacks**

N/A

**Applicable issues**

N/A

**Additional information**

**Checklist** 


.github/ct-install.yaml
```diff
@@ -5,6 +5,7 @@ chart-dirs:
 - charts/incubator
 - charts/stable
 excluded-charts:
+- charts/incubator/sdtd
```
bjw-s, line 8: Why did you exclude the chart from the install CI? Can it not run without specific hardware or are certain values required that you do not wish to set by default?

charts/incubator/sdtd/Chart.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: v2
+appVersion: 1.0.0
```
bjw-s, line 2:
```suggestion
appVersion: main
```

charts/incubator/sdtd/values.yaml
```diff
@@ -0,0 +1,62 @@
...
+#
+# This chart inherits from our common library chart. You can check the default values/options here:
+# https://github.com/k8s-at-home/library-charts/tree/main/charts/stable/common/values.yaml
+#
+
+image:
+  # -- image repository
+  repository: ghcr.io/reitermarkus/7d2d
+  # -- image tag
+  tag: main
```
bjw-s, line 12:
```suggestion
  # -- image tag
  # @default -- chart.appVersion
  tag:
```
This makes it so that we only have to update the appVersion (by default) but it still allows users to override it if they want.

charts/incubator/sdtd/values.yaml
```diff
@@ -0,0 +1,62 @@
...
+        enabled: true
+        port: 26900
+        protocol: UDP
+      steam:
+        enabled: true
+        port: 26901
+        protocol: UDP
+
+podSecurityContext:
+  fsGroup: 1000
```
bjw-s, line 46: We don't enable/set the podSecurityContext in our charts by default, unless we _really_ have to.

```suggestion
# -- Configure the Security Context for the Pod
podSecurityContext: {}
  # fsGroup: 1000  
```

charts/incubator/sdtd/values.yaml
```diff
@@ -0,0 +1,62 @@
...
+# @default -- See values.yaml
+persistence:
+  data:
+    enabled: false
+    mountPath: /data
+  server:
+    enabled: true
+    mountPath: /server
+    accessMode: ReadWriteOnce
+    size: 16Gi
```
bjw-s, line 58: We don't enable/set any persistence in our charts by default, unless we _really_ have to.

```suggestion
  server:
    enabled: false
    mountPath: /server
```

bjw-s (changes requested):

.github/ct-install.yaml
reitermarkus: It needs to download the whole game before it is `Ready`, which is about > 10 GB.

onedr0p (approved):

PR #1447, opened 2022-03-08
[otel-collector] add open telemetry collector with example config

mterhar:
**Description of the change**

Add a new service that acts as a local open telemetry collector and buffer. 

**Benefits**

Other apps within the k8s cluster can use this as an easy collection point to route telemetry to a backend of the user's choosing. 

**Possible drawbacks**

If people enable telemetry in other applications without knowing that this exists, it could be an administrative hassle. 

**Applicable issues**

N/A

**Additional information**

**Checklist** 


charts/stable/otel-collector/values.yaml
```diff
@@ -0,0 +1,179 @@
...
+#
+# This chart inherits from our common library chart. You can check the default values/options here:
+# https://github.com/k8s-at-home/library-charts/tree/main/charts/stable/common/values.yaml
+#
+
+image:
+  # -- image repository
+  repository: otel/opentelemetry-collector
+  # -- image tag
+  tag: 0.46.0
```
onedr0p, line 12: Remove "0.46.0", this will default to the version in the Chart.yaml

.all-contributorsrc
```diff
@@ -526,6 +526,16 @@
...
+    },
+
+    {
+      "login": "mterhar",
+      "name": "Mike Terhar",
+      "avatar_url": "https://avatars.githubusercontent.com/u/938684?v=4",
+      "profile": "http://www.brownfield.dev",
+      "contributions": [
+        "code"
+      ]
```
onedr0p, line 538: Revert this

onedr0p (changes requested):

charts/stable/otel-collector/Chart.yaml
```diff
@@ -0,0 +1,23 @@
...
+version: 1.0.0
+kubeVersion: ">=1.16.0-0"
+keywords:
+- otel-collector, open telemetry, tracing
+home: https://github.com/k8s-at-home/charts/tree/master/charts/stable/otel-collector
+icon: https://otel-collector.org/icon
+sources:
+- https://github.com/otel-collector/otel-collector-docker
+maintainers:
+- name: Mike Terhar
```
bjw-s, line 14: This is not a valid GitHub username, therefore the linting step is failing

charts/stable/otel-collector/templates/common.yaml
```diff
@@ -0,0 +1,20 @@
...
+type: "custom"
+volumeSpec:
+  secret:
+    {{- if .Values.configFileSecret }}
+    secretName: "{{ .Values.configSecretName }}"
+    {{- else }}
+    secretName: "{{ include "common.names.fullname" . }}-otelconfig"
+    {{- end }}
+{{- end -}}
+{{- $_ := set .Values.persistence "otel-config-file" (include "otel-collector.configVolume" . | fromYaml) -}}
```
bjw-s, line 18: Take a look at https://github.com/k8s-at-home/charts/blob/b1fc8dd547f6b9164e66824a661f0c2ff508ca9f/charts/stable/ghost/templates/common.yaml#L4L13 to see how we handle setting "hardcoded" values in our other charts.

charts/stable/otel-collector/templates/secret.yaml
```diff
@@ -0,0 +1,18 @@
+{{/*
+The open telemetry config secret to be included.
+*/}}
+{{- if and .Values.configFile (not .Values.configFileSecret) }}
+---
+apiVersion: v1
+kind: Secret
```
bjw-s, line 7: What is the reason for creating this as a Secret? I can see that there are potentially API keys being stored in here, but since they would be wide-open in the values.yaml already, there seems to be no real need for a Secret?

charts/stable/otel-collector/values.yaml
```diff
@@ -0,0 +1,179 @@
...
+  # -- image pull policy
+  pullPolicy: IfNotPresent
+
+# -- command is set to `otelcol` when using otel/opentelemetry-collector image but should be
+# changed to `otelcontribcol` if the image is changed to otel/opentelemetry-collector-contrib.
+# @default -- otelcol
+command:
+- "/otelcol"
```
bjw-s, line 20: This logic could be handled by the `hardcodedValues` solution as proposed in my comment on `common.yaml`

charts/stable/otel-collector/values.yaml
```diff
@@ -0,0 +1,179 @@
...
+
+# -- args is set to the config file which is mounted via the configFileSecret or configFile
+# @default -- Set to the hard-coded volume mount.
+args:
+- "--config=/conf/otel-collector-config.yaml"
```
bjw-s, line 25: This can be handled by the `hardcodedValues` solution as proposed in my comment on `common.yaml`

charts/stable/otel-collector/values.yaml
```diff
@@ -0,0 +1,179 @@
...
+      periodSeconds: 10
+      timeoutSeconds: 1
+      failureThreshold: 3
+      httpGet:
+        path: /
+        port: 13133
+  readiness:
+    enabled: false
+  startup:
+    enabled: false
```
bjw-s, line 78: What is the reasoning for omitting te readiness/startup probes?

charts/stable/otel-collector/values.yaml
```diff
@@ -0,0 +1,179 @@
...
+  # @default -- disabled
+  main:
+    enabled: false
+
+# -- Configure persistence settings for the chart under this key.
+# @default -- This application does not require persistence aside from the mounted secret
+persistence: {}
+  # data:
+  #   enabled: false
+  #   mountPath: /data
```
bjw-s, line 93: If the chart does not require persistence, this key can be removed from the yaml file

bjw-s (changes requested):

charts/stable/otel-collector/values.yaml
mterhar: There isn't a meaningful endpoint in the app for these and it starts super-fast since it's a go app. The health check should suffice.

charts/stable/otel-collector/values.yaml
mterhar: Which is preferable?

1. Assume everyone is okay with running the 50mb image with all the contrib plugins and just set the command to `/otelcontribcol` and the image to `otel-collector-contrib`. 
2. Expose a value that hints people into how to enable the contrib otel collector release

I think principle of least surprise is probably hard-coding the contrib release.

charts/stable/otel-collector/templates/secret.yaml
mterhar: I think assuming it's always a secret makes it easier to include the user-supplied one or the helm-configured one. In some situations it may not have any sensitive information in it and could be a configmap, but since that would require a lot of logic to support and doesn't benefit the user, I just leave it as a secret here. 

The admin can decide of they want the API key in a helm value yaml or if they want to handle that outside the chart, make a secret manually, and then reference it.

onedr0p (approved):

Issue #1455, opened 2022-03-18
wikijs helm chart's values.env overrides values.secrets


halkeye:
# Details

**Helm chart name and version:**

wikijs-6.2.0

**Container name and tag:**

ghcr.io/linuxserver/wikijs:version-2.5.201

**What steps did you take and what happened:**

Wanting to store database as a secret, I switched from env to secrets

**Relevant Helm values:**

```
secret:
  DB_HOST: postgresql.databases.svc.cluster.local
  DB_NAME: wikijs
  DB_TYPE: postgres
  DB_USER: wikijs
# plus actual secrets like passwords
```

**What did you expect to happen:**

pod's env is populated

**Anything else you would like to add:**

pod spec
```
  - env:
    - name: DB_FILEPATH
      value: /app/wiki/data/db.sqlite
    - name: DB_NAME
    - name: DB_PASS
    - name: DB_PASS_FILE
    - name: DB_PORT
    - name: DB_SSL
    - name: DB_SSL_CA
    - name: DB_TYPE
      value: sqlite
    - name: DB_USER
    - name: TZ
      value: America/Vancouver
    envFrom:
    - secretRef:
        name: wikijs
 
 ```

but when running export on the pod:

```
declare -x DB_FILEPATH="/app/wiki/data/db.sqlite"
declare -x DB_HOST="postgresql.databases.svc.cluster.local"
declare -x DB_NAME=""
declare -x DB_PASS=""
declare -x DB_PASS_FILE=""
declare -x DB_PORT=""
declare -x DB_SSL=""
declare -x DB_SSL_CA=""
declare -x DB_TYPE="sqlite"
declare -x DB_USER=""
```

Maybe common chart needs to be updated to exclude env that are also in the secret hash? or maybe the values for this chart need to not the keys pre-defined?

onedr0p:
I agree they should be at least commented out in the `values.yaml`, if you have a moment care to submit a PR?

onedr0p:
I think we have made this decision on other charts too since env and envFrom ordering matters on what takes precedence in the deployment spec

halkeye:
do you know offhand how commenting them out will affect the helm html/readme generation? Otherwise i'll go look at the docs when i poke at it tonight.

onedr0p:
I believe it will, maybe better to remove them and we can just link to the upstream docs if not doing so already

PR #1460, opened 2022-03-20
[monica] Remove default empty env variables - along the same lines as #1455

halkeye:
**Description of the change**

Remove default env variables for monica

**Benefits**

Allows env variables to be populated by secrets

**Possible drawbacks**

Anyone depending on the default values will have things break

**Applicable issues**

- related to #1455 (wikijs)

**Additional information**

**Checklist** 

### DIFF
--- a/charts/stable/monica/Chart.yaml
+++ b/charts/stable/monica/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 3.7.0-apache
 description: A Personal Relationship Management tool to help you organize your social life
 name: monica
-version: 7.0.1
+version: 7.0.2
 kubeVersion: ">=1.16.0-0"
 keywords:
   - crm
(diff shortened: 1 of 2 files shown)


onedr0p (approved):

samip5:
Why was this not an major version? That's a breaking change tho.

onedr0p:
How so? There were no default values changed, just envars commented out.

halkeye:
If i were to guess, APP_KEY might have accepted a blank string in the past? Its supposed to be a random hash, but maybe empty string was allowed? The "defaults" seem pretty bad to used

samip5:
> How so? There were no default values changed, just envars commented out.

If one used the env variables that were set in the helm chart, that's an breaking change to have them removed from the defaults even if such configuration is not secure.

halkeye:
> If one used the env variables that were set in the helm chart, that's an breaking change to have them removed from the defaults even if such configuration is not secure.

What defaults were you using that don't work now? The defaults were empty, so the new values are non set, in most languages they are the same.

Or is this just hypothetical?

Issue #1146, opened 2021-08-30
Home Assistant | Ingress need configuration


marcpaulchand:
# Details

**Helm chart name and version:**

k8s-at-home/home-assistant
version 10.0.0

**Container name and tag:**

homeassistant/home-assistant:2021.8.8

**What steps did you take and what happened:**

I edit the ingress section, in order to open access with the following yaml
```yml
ingress:
  # -- Enable and configure ingress settings for the chart under this key.
  # @default -- See values.yaml
  main:
    enabled: true
    ingressClassName: nginx
    hosts:
      - host: alfred.gotham #because i'm batman
        paths:
          - path: /
            pathType: Prefix
```

**What did you expect to happen:**

I should have access to the webUi, unfortunatly i was facing a simple `400 bad request`

The solution is pretty simple and need to adjust Home-assistant Configuration to accept our proxy.
But, it cannot be done through Docker environment variable as there is no ENV value in their dockerfile (not for that) [1] 
And i don't think relevant to think to edit `configuration.yml` with an other container at launch.

So in this king of situation, should the documentation mention this post-install adjustement ?
Or does users should find the solution by themselves ?

I think a simple edit in the README sending them to home assistant docs [2] with a little comment would be the kindest thing to do.

1: https://github.com/home-assistant/docker/blob/e61099e85be7fae319f2c5f281d8a945844b669e/Dockerfile
2: https://www.home-assistant.io/integrations/http

onedr0p:
Yea, this was a fun one to solve and it confuses many people. People need to somehow get to the config volume and edit the config to make it work.

kmcrawford:
The documentation on this is poor, as it looks to be something that should be added to the values.yaml, but it doesn't look like it.

onedr0p:
I disagree, Home assistant configuration is very opinionated and the file structure can vary between people's deployments.

PR #1463, opened 2022-03-22
#1146 specify editing the configuration.yaml

kmcrawford:
By stating that it's the configuration.yaml that needs to be modified, prevents users from thinking this is a values.yaml override.


bjw-s:
Incorporated in #1548

PR #1466, opened 2022-03-27
chore(homebridge): Set permissions for startup.sh to be executable

### LINKED ISSUE #1465: [homebridge] startup.sh is not mounted as executable
# Details

homebridge 4.3.0

**Container name and tag:**
name: homebridge
tag: `ghcr.io/oznu/homebridge:latest@sha256:4703dca68510307587361f4312e6bad0e765e85f8986a7517e0f393d8d0e2410`

Enable persistence and set a startup.sh script in the config stanza.
The application will boot as normal but silently fail because the upstream container expects to be able to `chmod +x`: https://github.com/oznu/docker-homebridge/blob/71e8eb26d4ee8eeb1adf0972b70e70ef261c3858/root/etc/cont-init.d/45-user-data#L11

**Relevant Helm values:**
```yaml
persistence:
  config:
    enabled: true
config: |
  #!/bin/sh
  echo "This message will not appear in the logs"
```

Startup script should execute

**Additional Information:**

seanson:
**Description of the change**

Currently we allow configuration of `startup.sh` through the `config` stanza in `values.yaml`, however this file is mounted with default permissions and so is not actually executable and fails silently. This adds a `defaultMode` configuration to the volume config which ensures the file will be executable at startup. 

**Benefits**

Ensures users running Homebridge will be able to configure and execute the startup script through the Helm configuration (commonly used for bootstrapping things like setting NET_CAP on the node binary).

**Possible drawbacks**

No drawbacks, as this fixes intended behaviour.

**Applicable issues**

- fixes #1465

**Additional information**

**Checklist**

### DIFF
--- a/charts/stable/homebridge/Chart.yaml
+++ b/charts/stable/homebridge/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 4.0.0
 description: A lightweight NodeJS server that emulates the iOS HomeKit API
 name: homebridge
-version: 4.3.0
+version: 4.3.1
 kubeVersion: ">=1.16.0-0"
 keywords:
   - homebridge
@@ -21,5 +21,5 @@ dependencies:
     version: 4.3.0
 annotations:
   artifacthub.io/changes: |
-    - kind: changed
-      description: Upgraded `common` chart dependency to version `4.3.0`.
+    - kind: fixed
+      description: Set permissions for startup.sh to be executable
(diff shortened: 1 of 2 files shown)


charts/stable/homebridge/Chart.yaml
```diff
@@ -22,4 +22,4 @@ dependencies:
 annotations:
   artifacthub.io/changes: |
     - kind: changed
```
bjw-s, line 24: Maybe change this to `fixed`

bjw-s (changes requested):
Thanks for the quick PR!! Only a tiny change requested :-)

charts/stable/homebridge/Chart.yaml
seanson: Done!

onedr0p (dismissed):

onedr0p (approved):

PR #1529, opened 2022-04-23
[uptime-kuma] Fixed bug in uptime kuma values.yaml regarding persistence

maxemann96:
**Description of the change**
Changed the key `persistence.data` to `persistence.config` in `values.yaml`

**Benefits**
Users without any prior knowledge of k8s@home charts are now easier be able to deploy this chart with persistence

**Possible drawbacks**
no

**Applicable issues**
no one

- fixes #

**Additional information**

**Checklist** 
- [ ] Variables have been documented in the `values.yaml` file. (Not needed, no change of actual variable)

### DIFF
--- a/charts/stable/uptime-kuma/Chart.yaml
+++ b/charts/stable/uptime-kuma/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 1.14.1
 description: A fancy self-hosted monitoring tool for your websites and applications
 name: uptime-kuma
-version: 1.2.2
+version: 1.2.3
 kubeVersion: ">=1.16.0-0"
 keywords:
   - uptime-kuma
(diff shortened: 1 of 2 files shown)


bjw-s:
I'm not entirely sure what the exact bug is here that you are trying to fix.

It's only the name of the volume that gets set within the Pod, and the mountPoint seems to indicate that it contains application data?

maxemann96:
As far as I understand the [persistence configuration](https://docs.k8s-at-home.com/our-helm-charts/common-library-storage/) of the common library used as dependency in this chart, it is needed to configurate the persistence under the key `persistence.config` and not `persistence.data` as stated in the `values.yaml` before.

maxemann96:
Got now how it's working. The initial error lead me to this change was that "accessMode is not defined for pvc ..." and I was searching for a solution why it not works. I found no documentation, that `persistence.x` is just a key defining the name of the pvc and was irritated, because the common chart template only documents the `config` key with some default values and I thought that there are then default values for all other keys than `config` or the key is wrong. 

Thanks for your patience

bjw-s:
No worries! Glad you got your initial issue sorted and thank you for wanting to submit a fix!

PR #1539, opened 2022-04-30
[paperles] Set PAPERLES_URL automatically

### LINKED ISSUE #1538: [paperless] CSRF verification failed
### Helm chart name

paperless

### Helm chart version

8.6.2

### Container name

ghcr.io/paperless-ngx/paperless-ngx:ngx-1.7.0

### Container tag

ngx-1.7.0

### Description

When installing Paperless with Ingress enabled, I get `Your CSRF verification failed` when I try to browse to the app

### Expected result

I expect to be able to access the app via the Ingress

### Helm values to reproduce

```YAML
ingress:
  main:
    enabled: true
    annotations:
      cert-manager.io/cluster-issuer: "letsencrypt-prod"
      nginx.ingress.kubernetes.io/proxy-body-size: "0"
      nginx.ingress.kubernetes.io/proxy-buffering: "off"
    ingressClassName: "public"
    hosts:
      - host: paperless.example.com
        paths:
          - path: /
            pathType: Prefix
    tls:
      - secretName: paperless-tls
        hosts:
          - paperless.example.com
```

### Additional Information

This seems to be discussed upstream in https://github.com/paperless-ngx/paperless-ngx/issues/712

The solution is to set the environment variable `PAPERLESS_URL` to the same value as the Ingress host.

In my case, setting

```yaml
env:
  PAPERLESS_URL: https://paperless.example.com
```

made the app work via the Ingress as expected, so I think there should be at least a placeholder for `PAPERLESS_URL` in `values.yaml`. Even better would be the ability to automatically set `PAPERLESS_URL` to the same value as `ingress.main.hosts.host[0]`

### Repo link

_No response_

angelnu:
**Description of the change**

Set PAPERLESS_URL by default to the main ingress host - see https://github.com/paperless-ngx/paperless-ngx/issues/712

**Benefits**

Working helm chart after upgrading to paperless-ngx >=v1.7

**Possible drawbacks**

None

**Applicable issues**

- fixes #1538 

**Additional information**

**Checklist** 

### DIFF
--- a/charts/stable/paperless/templates/common.yaml
+++ b/charts/stable/paperless/templates/common.yaml
@@ -1,1 +1,11 @@
+{{/* Make sure all variables are set properly */}}
+{{- include "common.values.setup" . }}
+
+{{/* Append the hardcoded settings */}}
+{{- define "paperless.harcodedValues" -}}
+env:
+    PAPERLESS_URL: {{ (first .Values.ingress.main.hosts).host }}
+{{- end -}}
+{{- $_ := merge .Values (include "paperless.harcodedValues" . | fromYaml) -}}
+
 {{ include "common.all" . }}
--- a/charts/stable/paperless/values.yaml
+++ b/charts/stable/paperless/values.yaml
@@ -36,7 +36,7 @@ env:
   # PAPERLESS_ADMIN_USER: admin
   # -- Password for the root user
   # PAPERLESS_ADMIN_PASSWORD: admin
-  # PAPERLESS_URL: <set do main ingress by default>
+  # PAPERLESS_URL: <set to main ingress by default>
 
 # -- Configures service settings for the chart.
 # @default -- See values.yaml
--- a/charts/stable/paperless/values.yaml
+++ b/charts/stable/paperless/values.yaml
@@ -36,6 +36,7 @@ env:
   # PAPERLESS_ADMIN_USER: admin
   # -- Password for the root user
   # PAPERLESS_ADMIN_PASSWORD: admin
+  # PAPERLESS_URL: <set do main ingress by default>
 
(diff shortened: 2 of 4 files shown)


charts/stable/paperless/values.yaml
```diff
@@ -36,6 +36,7 @@ env:
   # PAPERLESS_ADMIN_USER: admin
   # -- Password for the root user
   # PAPERLESS_ADMIN_PASSWORD: admin
+  # PAPERLESS_URL: <set do main ingress by default>
```
bjw-s, line 39:
```suggestion
  # PAPERLESS_URL: <set to main ingress by default>
```

charts/stable/paperless/templates/common.yaml
```diff
@@ -1 +1,11 @@
+{{/* Make sure all variables are set properly */}}
+{{- include "common.values.setup" . }}
+
+{{/* Append the hardcoded settings */}}
+{{- define "paperless.harcodedValues" -}}
+env:
+    PAPERLESS_URL: {{ (first .Values.ingress.main.hosts).host }}
```
bjw-s, line 7: Wouldn't doing it like this always override any custom values that have been set by the user?

bjw-s (changes requested):

charts/stable/paperless/templates/common.yaml
angelnu: I did a merge instead of mergeOverride to avoid overriding user values

angelnu: .. but override is not doing what I was expecting...

bjw-s (approved):

PR #1560, opened 2022-05-10
[onedrive] Update docker image to v2.4.17

wrmilling:
**Description of the change**

Bumping docker image version for OneDrive to v2.4.17. Docker image version now matches application version. 

**Benefits**

Latest code from upstream.

**Possible drawbacks**

N/A

**Applicable issues**

N/A

**Additional information**

N/A

**Checklist** 

### DIFF
--- a/charts/stable/onedrive/values.yaml
+++ b/charts/stable/onedrive/values.yaml
@@ -9,7 +9,7 @@ image:
   # -- image repository
   repository: ghcr.io/wrmilling/onedrive-docker
   # -- image tag
-  tag: v2.0.5
+  tag: v2.4.17
   # -- image pull policy
   pullPolicy: IfNotPresent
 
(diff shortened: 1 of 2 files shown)


charts/stable/onedrive/values.yaml
```diff
@@ -9,7 +9,7 @@ image:
   # -- image repository
   repository: ghcr.io/wrmilling/onedrive-docker
   # -- image tag
-  tag: v2.0.5
+  tag: v2.4.17
```
onedr0p, line 12:
```suggestion
  tag:
```

By default it will use the appVersion in the Chart.yaml, this makes it so we only need to manage the version in one place.

onedr0p (changes requested):

charts/stable/onedrive/values.yaml
wrmilling: Was not aware it would do that, neato.

onedr0p (approved):

PR #1562, opened 2022-05-12
[autobrr] Add autobrr chart

funkypenguin:
**Description of the change**

Hey guys! Love your work!

Here's a PR to add a chart for "autobrr" (https://github.com/autobrr/autobrr) - I hope I've done it right! 

**Benefits**

* New chart, happy users

**Possible drawbacks**

* None

**Applicable issues**

- fixes #

**Additional information**

**Checklist** 


charts/stable/autobrr/values.yaml
```diff
@@ -0,0 +1,87 @@
...
+#
+# This chart inherits from our common library chart. You can check the default values/options here:
+# https://github.com/k8s-at-home/library-charts/tree/main/charts/stable/common/values.yaml
+#
+
+image:
+  # -- image repository
+  repository: ghcr.io/autobrr/autobrr
+  # @default -- chart.appVersion
+  tag: v0.23.0
```
onedr0p, line 12:
```suggestion
  tag:
```

By default appVersion from the chart.yaml is used

onedr0p (changes requested):

onedr0p (approved):

onedr0p:
@all-contributors please add @funkypenguin for code

brettpetch:
Hey @onedr0p — it might be worth looking at this PR again. The app in question has a secret key that should be rotated across pods, I’m not seeing it being changed in the charts.

onedr0p:
Huh? Sorry I'm not at all familiar with this application.

charts/stable/autobrr/values.yaml
```diff
@@ -0,0 +1,87 @@
...
+  # Default: "DEBUG"
+  #
+  # Options: "ERROR", "DEBUG", "INFO", "WARN"
+  #
+  logLevel = "INFO"
+
+  # Session secret
+  #
+  sessionSecret = "secret-session-key"
+
```
brettpetch, line 64: Secret key should be rotated.

brettpetch: Example of session key: https://github.com/swizzin/swizzin/blob/3e29425507ef6711cd36791c008490ecf32689df/scripts/install/autobrr.sh#L52

charts/stable/autobrr/README.md
```diff
@@ -0,0 +1,104 @@
+# autobrr
+
+![Version: 13.4.3](https://img.shields.io/badge/Version-13.4.3-informational?style=flat-square) ![AppVersion: v4.4.2](https://img.shields.io/badge/AppVersion-v4.4.2-informational?style=flat-square)
+
+autobrr is a cross-platform free and open-source BitTorrent client
```
brettpetch, line 5: This should actually read 

> Autobrr monitors IRC announce channels to get releases as soon as they are available with good filtering. Built on Go/React to be resource friendly. A modern single binary replacement for autodl-irssi+rutorrent plugin.

onedr0p:
@funkypenguin what do you think of these changes?

charts/stable/autobrr/values.yaml
```diff
@@ -0,0 +1,87 @@
...
+  # Custom definitions
+  #
+  #customDefinitions = "test/definitions"
```
brettpetch, line 67: might be worth swapping this to just being enabled by default in the config folder somewhere, maybe /config/definitions

funkypenguin:
The app description was a copy/paste oversight, I'll submit a PR to fix that :)

WRT the session secret, my feeling is that in this context, we're better served by keeping the helm chart simple, and avoiding auto-created config values. The chart user is free to alter the config their `values.yaml`, and set the session secret themselves, but doing it programmatically would be fiddly, and not worth it IMO.

If the config options could also be passed to the app via ENV vars (_which override the config_), then there's  could be a case made for further automating the configuration.

brettpetch:
Allowing users who have successfully authenticated on another instance with the same key could just login to this one using the same cookie... How is that a good solution to authentication? The session key should be generated for every pod as it pretty much allows unfettered access to the person's accounts, api keys, filters, download clients, and is a huge RCE vulnerability in itself if left unchecked.

PR #1566, opened 2022-05-13
[autobrr] Update app description

funkypenguin:
Hey guys,

Just a minor update to correct an oversight during copy/pastage. The app description is now correct :)

D


onedr0p:
Looks like there's some whitespace to be chomped

onedr0p (approved):

PR #1586, opened 2022-05-28
[send] Added timvisee Send chart

geraldwuhoo:
Signed-off-by: geraldwuhoo <gerald@geraldwu.com>

**Description of the change**

Add [Send](https://gitlab.com/timvisee/send) chart

**Benefits**

People will be able to deploy Send, a simple drag-and-drop file sharing service with client-side encryption.

**Possible drawbacks**

NA

**Applicable issues**

NA

**Additional information**

**Checklist** 


charts/stable/send/values.yaml
```diff
@@ -0,0 +1,48 @@
...
+# This chart inherits from our common library chart. You can check the default values/options here:
+# https://github.com/k8s-at-home/library-charts/tree/main/charts/stable/common/values.yaml
+#
+
+image:
+  # -- image repository
+  repository: registry.gitlab.com/timvisee/send
+  # -- image tag
+  # @default -- chart.appVersion
+  tag: v3.4.20
```
bjw-s, line 13:
```suggestion
  tag:
```

bjw-s (changes requested):
Thanks for the PR! Just a tiny change requested :)

geraldwuhoo:
Thanks! For future reference, why not include the tag in the base `values.yaml`?

charts/stable/send/values.yaml
geraldwuhoo: Done

bjw-s:
Thank you for updating! 

> Thanks! For future reference, why not include the tag in the base `values.yaml`?

The reason for this is that the value gets populated by the `appVersion` field from `Chart.yaml` metadata by default when the field is empty. We had quite some charts in the past where the `appVersion` metadata would differ from the `image.tag` value. So by leaving it empty by default we ensure that they are always in sync (and only have to update the application version in one spot) :) Users can still manually override/pin the value by specifying a value themselves.

bjw-s (approved):

bjw-s:
@all-contributors Please add @geraldwuhoo for code

PR #1587, opened 2022-05-29
[mealie] Support mealie v1.0.0

Jonnobrow:
**Description of the change**

Adds support for mealie v1 (currently in beta/nightly).

**Benefits**

Lets users use all the features from the new version of mealie.

**Possible drawbacks**

Completely incompatible with all previous versions and currently in beta.

**Applicable issues**

- N/A

**Additional information**

**Checklist** 

### DIFF
--- a/charts/stable/mealie/Chart.yaml
+++ b/charts/stable/mealie/Chart.yaml
@@ -1,9 +1,9 @@
 ---
 apiVersion: v2
-appVersion: v0.5.6
+appVersion: v1.0.0-beta-1
 description: Mealie is a self hosted recipe manager and meal planner with a RestAPI backend and a reactive frontend application built in Vue for a pleasant user experience for the whole family.
 name: mealie
-version: 4.2.2
+version: 5.0.0
 kubeVersion: ">=1.16.0-0"
 keywords:
   - grocy
@@ -14,15 +14,15 @@ sources:
 maintainers:
   - name: bjw-s
     email: me@bjw-s.dev
+  - name: jonnobrow
+    email: jonathan@jonnobrow.co.uk
 dependencies:
   - name: common
     repository: https://library-charts.k8s-at-home.com
     version: 4.4.2
-  - name: postgresql
-    repository: https://charts.bitnami.com/bitnami
-    version: 10.16.2
-    condition: postgresql.enabled
 annotations:
   artifacthub.io/changes: |-
     - kind: changed
-      description: Upgraded `common` chart dependency to version 4.4.2
+      description: Removed `postgresql` chart dependency.
+    - kind: changed
+      description: Updated mealie to v1.0.0-beta-1.
--- a/charts/stable/mealie/Chart.yaml
+++ b/charts/stable/mealie/Chart.yaml
@@ -1,6 +1,6 @@
 ---
 apiVersion: v2
-appVersion: v1.0.0-beta-1
+appVersion: v1.0.0beta-2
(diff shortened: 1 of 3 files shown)


bjw-s:
I'm not too sure we want to be running beta software by default if I'm honest. @k8s-at-home/maintainers how do you feel about this?

angelnu:
As it is following the upstream project then I think we should follow as well - using a new mayor version allows people to stay in the v0.x if they want to.

Alternative would be to create a second chart mealie-beta instead of updating the current one but I think this is overkill.

And yes, sure there will be people upgrading to a new mayor and complaining that they need to update their config :-)

Truxnell:
I agree with @angelnu - It sounds like bumping major semver to follow upstream is best.  Not keen on a beta chart.
We will want to put a strong callout on the bump but we will have to accept some level of 'I just upgraded and it broke'

charts/stable/mealie/Chart.yaml
```diff
@@ -14,15 +14,15 @@ sources:
...
-    repository: https://charts.bitnami.com/bitnami
-    version: 10.16.2
-    condition: postgresql.enabled
 annotations:
   artifacthub.io/changes: |-
     - kind: changed
-      description: Upgraded `common` chart dependency to version 4.4.2
+      description: Removed `postgresql` chart dependency.
+    - kind: changed
+      description: Updated mealie to v1.0.0-beta-1.
```
bjw-s, line 28:
```suggestion
      description: "**BREAKING**: Updated mealie to v1.0.0-beta-1. Please check the application documentation for upgrade steps."
```

charts/stable/mealie/values.yaml
```diff
@@ -1,34 +1,28 @@
-#
 # IMPORTANT NOTE
 #
 # This chart inherits from our common library chart. You can check the default values/options here:
 # https://github.com/k8s-at-home/library-charts/tree/main/charts/stable/common/values.yaml
 #
-
+---
```
bjw-s, line 6: Can you please revert the header changes? We want to keep this the same throughout all charts.

charts/stable/mealie/values.yaml
```diff
@@ -1,34 +1,28 @@
...
 image:
   # -- image repository
   repository: hkotel/mealie
   # -- image tag
   # @default -- chart.appVersion
-  tag:
+  tag: frontend-v1.0.0beta-2
```
bjw-s, line 12: The `appVersion` field in Chart.yaml refers to version v1.0.0-beta-1, should they not be in sync?

charts/stable/mealie/values.yaml
```diff
@@ -1,34 +1,28 @@
...
   pullPolicy: IfNotPresent
 
-# -- environment variables. See [image docs](https://hay-kot.github.io/mealie/documentation/getting-started/install/#env-variables) for more details.
-# @default -- See below
-env:
-  # -- Set the container timezone
-  TZ: UTC
-  # -- Set the application database type
-  DB_ENGINE: sqlite
-
```
bjw-s, line 24: There still seem to be quite some env vars according to the docs. Please don't completely remove this section. Perhaps only remove the `DB_ENGINE` var?

bjw-s (changes requested):

charts/stable/mealie/values.yaml
Jonnobrow: Since there is a tag prefix we can't use chart.appVersion by default. I have made app version match the suffix though.

bjw-s (approved):

PR #1622, opened 2022-06-19
[adguard-home] Remove version tag

HenrikBacher:
Remove version specific to get latest version

Remove the version tag in Values.yaml, so latest version of adguard-home will be fetch by default

**Additional information**

**Checklist** 

### DIFF
--- a/charts/stable/adguard-home/values.yaml
+++ b/charts/stable/adguard-home/values.yaml
@@ -37,7 +37,7 @@ image:
   # -- image repository
   repository: adguard/adguardhome
   # -- image tag
-  tag: v0.106.3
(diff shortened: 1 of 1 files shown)


bjw-s:
Thanks for creating the PR!. I'm afraid we can't merge this change though. 
By design we do not want our charts running on the `latest` tag if we can help it. Running a specific version of an application  is a lot easier to maintain and manage, since versions should be immutable and you therefore will always deploy the same "state".

If for some reason you do wish to run the `latest` in your cluster, you can always override the variable in your `values.yaml` (or it's equivalent for the way you manage your Helm releases)

PR #1632, opened 2022-06-25
[omada-controller] Provide useful defaults

dimakuz:
**Description of the change**

Added values for:
* Data volume mount
* Ports required to manage EAP access points and other TP Link devices
* Commented out annotations for nginx-ingress to get ingress working.

**Benefits**

Simpler set up

**Possible drawbacks**

N/A

### DIFF
--- a/charts/stable/omada-controller/values.yaml
+++ b/charts/stable/omada-controller/values.yaml
@@ -26,15 +26,48 @@ service:
     ports:
       http:
         port: 8043
+  management:
+    enabled: false
+    ports:
+      tcp29811:
+        enabled: true
+        port: 29811
+        protocol: TCP
+        primary: false
+      tcp29812:
+        enabled: true
+        port: 29812
+        protocol: TCP
+        primary: false
+      tcp29813:
+        enabled: true
+        port: 29813
+        protocol: TCP
+        primary: false
+    discovery:
+      enabled: false
+      ports:
+        udp29810:
+          enabled: true
+          port: 29810
+          protocol: UDP
+          primary: false
 
 ingress:
   # -- Enable and configure ingress settings for the chart under this key.
   # @default -- See values.yaml
   main:
     enabled: false
+    # annotations:
+    #   nginx.ingress.kubernetes.io/backend-protocol: HTTPS
+    #   nginx.ingress.kubernetes.io/proxy-redirect-from: https://$http_host:8043/
+    #   nginx.ingress.kubernetes.io/proxy-redirect-to: https://$http_host/
+    #   nginx.ingress.kubernetes.io/configuration-snippet: |
+    #     proxy_set_header        Host $http_host:8043;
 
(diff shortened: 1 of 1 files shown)


Truxnell:
Hi @dimakuz, thanks for the PR!

Looks like you haven't updated the changelog in the `Chart.yaml`, can you please do so?

charts/stable/omada-controller/values.yaml
```diff
@@ -26,15 +26,48 @@ service:
     ports:
       http:
         port: 8043
+  management:
```
bjw-s, line 29: I don't think it really adds much value to run three separate Services for this. Could you please take a look at how the Unifi Controller chart (https://github.com/k8s-at-home/charts/blob/master/charts/stable/unifi/values.yaml#L47) handles this, and implement it in a similar way?

charts/stable/omada-controller/values.yaml
```diff
@@ -26,15 +26,48 @@ service:
...
+    enabled: false
+    ports:
+      tcp29811:
```
bjw-s, line 32: Would it be possible to add descriptive names to these ports?

charts/stable/omada-controller/values.yaml
```diff
@@ -26,15 +26,48 @@ service:
...
   # -- Enable and configure ingress settings for the chart under this key.
   # @default -- See values.yaml
   main:
     enabled: false
+    # annotations:
+    #   nginx.ingress.kubernetes.io/backend-protocol: HTTPS
+    #   nginx.ingress.kubernetes.io/proxy-redirect-from: https://$http_host:8043/
+    #   nginx.ingress.kubernetes.io/proxy-redirect-to: https://$http_host/
+    #   nginx.ingress.kubernetes.io/configuration-snippet: |
+    #     proxy_set_header        Host $http_host:8043;
```
bjw-s, line 66: We try to keep our charts Ingress-agnostic.
Even though the annotations are commented out, it would be preferred if you would describe the required configuration in the `README_CONFIG.md.gotmpl` file. 

I don't know the Omada controller well enough, but I'm also not sure all 4 of these are 100% required? I can imagine that it depends on how the rest of the Ingress (controller) is set up.

charts/stable/omada-controller/values.yaml
```diff
@@ -26,15 +26,48 @@ service:
...
 
 # -- Configure persistence settings for the chart under this key.
 # @default -- See values.yaml
 persistence:
   data:
     enabled: false
+    mountPath: /opt/tplink/EAPController/data
```
bjw-s, line 73: Good catch! 👍

bjw-s (changes requested):

charts/stable/omada-controller/values.yaml
legonigel: FYI, 29811, 29812, 29813 are all device upgrade/management ports.

PR #1710, opened 2022-07-26
[mosquitto] Remove storage fields from configinc value

terop:
**Description of the change**

Persistent storage settings for `configinc` have been removed because they should not have been there in the first place as they are to be specified by the user.

**Benefits**

The chart is aligned to the best practices used in this repository.

**Possible drawbacks**

The user must now specify the storage manually.

**Checklist** 
- [ X] Title of the PR starts with chart name (e.g. `[home-assistant]`)
- [ X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [ X] Chart `artifacthub.io/changes` changelog annotation has been updated in `Chart.yaml`. See [Artifact Hub documentation](https://artifacthub.io/docs/topics/annotations/helm/#supported-annotations) for more info.
- [ X] Variables have been documented in the `values.yaml` file.

### DIFF
--- a/charts/stable/mosquitto/Chart.yaml
+++ b/charts/stable/mosquitto/Chart.yaml
@@ -3,7 +3,7 @@ apiVersion: v2
 appVersion: 2.0.14
 description: Eclipse Mosquitto - An open source MQTT broker
 name: mosquitto
-version: 4.7.0
+version: 4.8.0
 kubeVersion: ">=1.16.0-0"
 keywords:
   - mosquitto
@@ -23,4 +23,4 @@ dependencies:
 annotations:
   artifacthub.io/changes: |-
     - kind: changed
-      description: Upgraded `common` chart dependency to version 4.5.0
+      description: Remove some storage fields from the `persistence.configinc` value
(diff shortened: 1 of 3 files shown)


bjw-s:
Hi! Thanks for taking the time to raise this PR! Unfortunately it has exposed something in the chart that should not have been there in the first place. We try to keep the `persistence` fields as minimal as possible and let the user determine their desired values based on their needs (guided by http://docs.k8s-at-home.com/our-helm-charts/common-library-storage/).

That means we will / should be removing the following fields:

```
persistence:
  configinc:
    accessMode: ReadWriteOnce
    # The default storage class is used when the null value is used. Set this
    # to non-null to use a different storage class.
    storageClass: null
    size: 100Mi
```

terop:
Okay, thanks for your input @bjw-s. Should I close this PR or update it to remove the part you indicated?

bjw-s:
> Okay, thanks for your input @bjw-s. Should I close this PR or update it to remove the part you indicated?

It would be great if you could update the PR to remove it. Thanks so much!!

terop:
PR updated to remove the fields.

bjw-s (approved):

PR #1711, opened 2022-07-26
[mosquitto] Add option for selectively adding listener

### LINKED ISSUE #1701: [mosquitto] TLS does not work if listener is defined in mosquitto.conf
### Helm chart name

mosquitto

### Helm chart version

4.4.0

### Container name

eclipse-mosquitto

### Container tag

2.0.14

### Description

If TLS is to be used (which nowadays is preferred, I suppose) defining the `listener` option in the main `mosquitto.conf` causes TLS authentication to not work, below lines are seen in the log
```
1657982972: New connection from <IP>:<port> on port 8883.
1657982972: Client <unknown> disconnected due to malformed packet.
```

 TLS works if  the`listener` option is defined in the same config file where the TLS `certfile`and `keyfile` options are defined. This problem has noted in other places, e.g. https://serverfault.com/a/1075260/975505. I guess TLS would also work if the certificate options were defined in `mosquitto.conf` but I think it is not recommended by upstream.

I workarounded this problem by manually modifying the chart and removing the `listener` option from `configmap.yaml`, see diff below.
```
index a2bba09e..e389901a 100644
--- a/charts/stable/mosquitto/templates/configmap.yaml
+++ b/charts/stable/mosquitto/templates/configmap.yaml
@@ -9,7 +9,6 @@ metadata:
 data:
   mosquitto.conf: |
     per_listener_settings {{ .Values.perListenerSettings }}
-    listener {{ .Values.service.main.ports.mqtt.port }}
     {{- if .Values.auth.enabled }}
     allow_anonymous false
     {{- else }}
```

To actually solve the problem I see two options:
1) Add a note in the chart documentation that if you want to use TLS the `listener` option needs to be removed
or
2) Add a new value to `values.yaml` which would enable conditionally to do 1) above without the user having to modify the chart

I think implementing option 2) would make sense.

### Expected result

TLS works without having to modify the chart.

### Helm values to reproduce

```YAML
image:
  tag: 2.0.14

service:
  main:
    ports:
      mqtt:
        enabled: true
        port: 8883

auth:
  enabled: true

persistence:
  configinc:
    enabled: true
    mountPath: /mosquitto/configinc
    accessMode: ReadWriteOnce
    size: 50Mi
```

### Additional Information

_No response_

### Repo link

_No response_

terop:
**Description of the change**

Add an option to control adding of the `listener` option to mosquitto.conf. Using TLS in Mosquitto does not work if the `listener` option is present in the main mosquitto.conf. The linked issue has a longer description on the problem.

**Benefits**

TLS can be configured without having to manually edit the chart.

**Possible drawbacks**

None, the change is fully backwards compatible.

**Applicable issues**

- Fixes #1701

**Checklist** 
- [ X] Title of the PR starts with chart name (e.g. `[home-assistant]`)
- [ X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [ X] Chart `artifacthub.io/changes` changelog annotation has been updated in `Chart.yaml`. See [Artifact Hub documentation](https://artifacthub.io/docs/topics/annotations/helm/#supported-annotations) for more info.
- [X ] Variables have been documented in the `values.yaml` file.

### DIFF
--- a/charts/stable/mosquitto/values.yaml
+++ b/charts/stable/mosquitto/values.yaml
@@ -32,6 +32,10 @@ auth:
 # -- By enabling this, authentication and access control settings will be controlled on a per-listener basis
 perListenerSettings: false
 
+# -- When enabled, this adds the `listener` option to the mosquitto config.
+# Change this to false when using TLS.
+addListener: true
+
 persistence:
   # -- Configure a persistent volume to place mosquitto data in.
   # When enabled, this enables `persistence` and `persistence_location` in the mosquitto config.
--- a/charts/stable/mosquitto/Chart.yaml
+++ b/charts/stable/mosquitto/Chart.yaml
@@ -3,7 +3,7 @@ apiVersion: v2
 appVersion: 2.0.14
 description: Eclipse Mosquitto - An open source MQTT broker
 name: mosquitto
-version: 4.5.1
+version: 4.6.0
 kubeVersion: ">=1.16.0-0"
 keywords:
   - mosquitto
@@ -22,5 +22,5 @@ dependencies:
     version: 4.4.2
 annotations:
   artifacthub.io/changes: |-
-    - kind: changed
-      description: Use appVersion as image tag by default
+    - kind: added
+      description: Option to disable adding of the listener option to the config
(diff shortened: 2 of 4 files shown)


charts/stable/mosquitto/values.yaml
```diff
@@ -18,6 +18,9 @@ image:
 # @default -- See values.yaml
 service:
   main:
+    # When enabled, adds the `listener` option to the mosquitto config. Change
+    # this to false when using TLS.
+    addListener: true
```
bjw-s, line 23: I don't think this should live under the `service.main` key. Please place this on the values root, underneath `perListenerSettings: false`.

bjw-s (changes requested):

bjw-s (approved):
Thanks very much for the Pull Request! Approved and merged :)

